PR #26, opened 2021-04-11
Lua scripting framework

SeanReg:
Basic scripting framework that allows Users to create an "OswLuaApp" which points to a Lua script. Lua script is then loaded and executed as an app.

1. Scripts are currently stored as static strings in a .h file (using PROGMEM to store in flash). GPS edition could load from SD. For Light edition, probably should have build process that converts a .lua to .h
2. ~Lua script does not currently get a pointer to the Hal or the Canvas. Lua bindings (maybe via SWIG) need to be created to allow Lua scripts to draw to the canvas~
2. **Update** Added SWIG interface to generate bindings for OswHal as well as Graphics2d. This allows Lua code to call into the OswHal (see example). Should be able to use this to draw to the screen from Lua

Following can be used to regenerate the _osw_wrap.cxx_ This should maybe be automated as part of the build process?
`swig -c++ -lua -I../../include -I../../lib/Arduino_GFX osw.i`

Example script currently just prints to stdio and looks like this:

```
function setup()

end

function loop()
    if (hal:btn1Down() > 0) then
        print("btn1 pressed")
    else
        print("btn1 not pressed")
    end
end

function stop()

end
```

```
RAM:   [=         ]  13.6% (used 44488 bytes from 327680 bytes)
Flash: [=====     ]  49.3% (used 1033765 bytes from 2097152 bytes)
```


uvwxy:
Perfect!

Issue #30, opened 2021-04-13
Brightness


max9901:
increase / decrease brightness hal functions don't actually do anything.

PR #31, opened 2021-04-13
fix for issue #30 brightness

max9901:


src/hal/display.cpp
```diff
@@ -90,11 +90,21 @@ void OswHal::increaseBrightness(uint8_t v) {
   } else {
     _brightness += v;
   }
+  #ifdef ESP32
```
uvwxy, line 93: could you change this to setBrightness(_brightness)

would be less code duplication that way. thank you :)

src/hal/display.cpp
```diff
@@ -90,11 +90,21 @@ void OswHal::increaseBrightness(uint8_t v) {
...
+    digitalWrite(TFT_LED, brightness);
+  #endif
 };
 void OswHal::decreaseBrightness(uint8_t v) {
   if (_brightness < v) {
     _brightness = 0;
   } else {
     _brightness -= v;
   }
+  #ifdef ESP32
```
uvwxy, line 105: see above 👆

uvwxy (changes requested):
If you change the setting of the brightness to use the existing setter function, I can merge this

src/hal/display.cpp
max9901: Yeah but its 1 more memory operation. so slower.

max9901: But yeah if this is "cleaner" in your opinion i can change it ofcourse!

uvwxy: I understand you.

My perspective was: fewer places to change stuff if we eventually use a different display.
Also 1 more memory operation versus 240Mhz on a function that currently get's called once when the watch wakes up and once when it goes back to sleep. Even gradually dimming at 100fps would still leave tons of overhead of performance...

uvwxy (approved):

Issue #29, opened 2021-04-13
Auto-Build osw_wrap.cxx


uvwxy:
See: https://github.com/Open-Smartwatch/open-smartwatch-os/pull/26

> Following can be used to regenerate the osw_wrap.cxx This should maybe be automated as part of the build process?
swig -c++ -lua -I../../include -I../../lib/Arduino_GFX osw.i

uvwxy:
@Apfelwurm this might be something for you, if you have the time

Thanks in advance :)

Apfelwurm:
Is now on my todo list :)
This should be easily possible via a prebuild/postbuild script within pio.

Apfelwurm:
I think this one could be closed :)

PR #33, opened 2021-04-17
implemented swig build and updated ci and readme

Apfelwurm:
swig building implementation thats needed for https://github.com/Open-Smartwatch/open-smartwatch-os/issues/29

Can someone maybe test it on windows? i have currently no access to a running windows machine :D 
I'm not sure if the paths also work on windows like this is built. If there are any problems i can also implement a additional switch for that.


Apfelwurm:
@SeanReg reported that the build also works on windows, so i think we are good to go :)

PR #34, opened 2021-04-19
Lua Support for "Print" methods

SeanReg:
1. Support for Canvas "Print" methods in Lua 
2. Remove ignore osw_wrap from commits and gitignroe

@Apfelwurm
 Modified the pre-build script. Let me know if you see any issues. Print.h is part of the Framework include files so I had to try to come up with a way to solve its path from the Pio Env.


Apfelwurm:
looks good to me :) good to see that my thought that it might be a good idea to implement the additional buildstep with the pre script in pio pays off. And i like the verbose option, its much better to see whats going on.

PR #40, opened 2021-04-23
Reset step count each day at midnight

mmoneyron:
This feature reset step count each day so that you can track only today's activity


src/main.cpp
```diff
@@ -117,6 +117,14 @@ void setup() {
 
 void loop() {
   static long lastFlush = 0;
+  uint32_t second = 0;
+  uint32_t minute = 0;
+  uint32_t hour = 0;
+
+  hal->getLocalTime(&hour, &minute, &second);
+  if ((hour == 0) && (minute == 0) && (second == 0)) {
```
uvwxy, line 125: the watch will be in deep sleep most of the time, especially at night.

I like the idea, but it won't work that easy.

We have to look into waking up the system via an RTC interrupt and check on demand to reset the step count.

mmoneyron: Yes you're right. For testing purpose I disabled auto sleep, so I didn't had that problem.
I'll look at it

mmoneyron:
This new commit makes the watch to enter deep-sleep with a timer which will wakeup at xxh00 and xxh30, then re-enable the deep-sleep for 30min.
This way the step count can be reset at midnight.

However, it's a bit hacky because when the watch wakeup from timer, the screen is setup for a second before going back to sleep. So this emits light every half an hour, even at night.

uvwxy:
I think we are getting closer here to a solution. At the same time I'm thinking about broadening the scope of resetting the steps and adding them to a histogram of steps... 🤔 

Maybe discuss this on discord?

rasppliances:
> This new commit makes the watch to enter deep-sleep with a timer which will wakeup at xxh00 and xxh30, then re-enable the deep-sleep for 30min.
> This way the step count can be reset at midnight.
> 
> However, it's a bit hacky because when the watch wakeup from timer, the screen is setup for a second before going back to sleep. So this emits light every half an hour, even at night.

What if the user is walking at midnight?  Or the watch is on at midnight?  I'm not sure where in the code the steps are counted when the watch is in a deep sleep, but I think that might be the place to check the time and reset the steps.

mmoneyron:
Well, from my understanding, the steps are counted directly in the BMA400 hardware device. And the code only reads the steps count from the BMA400 when not in deep sleep (updateAccelerometer() in lhe main loop). Since the counter is not just a variable, but it's a hardware counter, to reset steps count my code soft resets the BMA400 then set it up again like at first power on.

If the user is using the watch at midnight, there is no problem since the check is done in the main loop.

rasppliances:
I don't think that resetting the chip as a whole is a good idea.  There are other things that chip does that could be impacted.  For example, the twist wrist to turn on code I submitted relies on certain registers being setup a certain way for the interrupt to work.

But, main.cpp runs from the start on every power on, which means sensors.cpp is called on every startup, which means my code is run on every setup, so really it shouldn't be impacted.

I still think it's a better idea to only reset what you need to reset, and not the entire thing.

mmoneyron:
Indeed you're right. It is better to only reset what's needed. I'll try to figure it out.

rasppliances:
@mmoneyron I hope you don't think I was being nasty?  I'm sorry if I was. I'm also sorry if I misread your last comment.

mmoneyron:
Oh no, don't worry about that :). I'm totally open to improvement suggestions

mmoneyron:
Now only the step counter register is reset.

There is one issue remaining concerning screen backlight on wakeup from timer every 30 minutes. After some investigation, I think this could be addressed with something like the stub [esp_wake_deep_sleep](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/sleep_modes.html#_CPPv419esp_wake_deep_sleepv). However I didn't managed to make it work.

richtepa:
> There is one issue remaining concerning screen backlight on wakeup from timer every 30 minutes.

I think the backlight is enabled in [main.cpp:110](https://github.com/Open-Smartwatch/open-smartwatch-os/blob/8e0c7daee4a2059e6b26a06385b10c7831f83c5a/src/main.cpp#L110): ` hal->setBrightness(128);`
I'm not sure about timing but maybe you can check for `hal->isWakeupFromTimer()` right before and skip lighting the screen up.

mmoneyron:
@richtepa I aleady tried what you suggest, but the problem is still here. I put a delay in the setup function, just before `Serial.begin(115200)`. And there is still backlight.
I guess the underlying OS is powering the screen before the setup function. That's why I suggest to run the stub on wakeup before bootloader or osw app, which would check time, reset steps count if needed, then go back to deep sleep.

richtepa:
@mmoneyron I just checked it with my watch.

> but the problem is still here

You are (obviously) right.

> I think this could be addressed with something like the stub esp_wake_deep_sleep. However I didn't managed to make it work.

Your idea sounds good. Sadly I cannot help you getting this to work.

uvwxy:
I will implement something for this, but this depends on my progress on other topics.

PR #61, opened 2021-05-04
Added troubleshooting section to readme and extra commenting to config example

Sharealikelicence:
Added troubleshooting section to readme and added some extra commenting to config example.


uvwxy:
Thank you!

PR #62, opened 2021-05-04
Update watchface_digital.cpp

rawsignal:
Correcting an issue where if the step count was too low (or presumably too high) the centering would be incorrect. It should now be dynamic and tested up to a step count of 5 digits, as low as 1, to ensure centering works as expected.

### DIFF
--- a/src/apps/main/watchface_digital.cpp
+++ b/src/apps/main/watchface_digital.cpp
@@ -14,17 +14,6 @@ using std::string;
 
 #define COLOR_BLACK rgb565(0, 0, 0)
 
-int numPositiveDigits(uint32_t x)
-{
-    if (x < 10) return 1;
-    if (x < 100) return 2;
-    if (x < 1000) return 3;
-    if (x < 10000) return 4;
-    if (x < 100000) return 5;
-    if (x < 1000000) return 6;
-    if (x < 10000000) return 7;
-}
-
 void drawDate(OswHal* hal) {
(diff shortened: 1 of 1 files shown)


rawsignal:
numPositiveDigits could be useful as a helper somewhere, not sure where we could put it though

src/apps/main/watchface_digital.cpp
```diff
@@ -90,8 +101,9 @@ void drawTime24Hour(OswHal* hal) {
 
 void drawSteps(OswHal* hal) {
   uint32_t steps = hal->getStepCount();
+  int width = numPositiveDigits(steps);
   hal->gfx()->setTextSize(2);
-  hal->gfx()->setTextCursor(((220/2)-defaultFontXOffset(1, 2)), 210 - defaultFontYOffset(1, 2) / 2);
+  hal->gfx()->setTextCursor(120 - hal->gfx()->getTextOfsetColumns(width/2), 210 - defaultFontYOffset(1, 2) / 2);
```
uvwxy, line 106: better: `setTextCenterAligned()`, then we don't need the numPositiveDigits.

uvwxy (changes requested):

src/apps/main/watchface_digital.cpp
rawsignal: Updated to use setTextCenterAligned and tested working great

PR #63, opened 2021-05-05
configurable date

rawsignal:
This will allow the date display (currently only available on the digital watchface) to be configured one of two ways

### DIFF
--- a/src/apps/main/watchface_digital.cpp
+++ b/src/apps/main/watchface_digital.cpp
@@ -35,14 +35,23 @@ void drawDate(OswHal* hal) {
 
     hal->gfx()->print(date_Array);
     hal->gfx()->print(" ");
-    hal->gfx()->printDecimal(monthInt, 2);
-    hal->gfx()->print("/");
-    hal->gfx()->printDecimal(dayInt, 2);
+
+    //i really would want the date to be dynamic based on what's in the config, but the most efficient thing to do right now is simply two if statements covering the 2 common conditions.
+    if (DATE_FORMAT == "mm/dd/yyyy") {
+      hal->gfx()->printDecimal(monthInt, 2);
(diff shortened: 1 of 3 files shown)


src/apps/main/watchface_digital.cpp
```diff
@@ -35,14 +35,23 @@ void drawDate(OswHal* hal) {
 
     hal->gfx()->print(date_Array);
     hal->gfx()->print(" ");
-    hal->gfx()->printDecimal(monthInt, 2);
-    hal->gfx()->print("/");
-    hal->gfx()->printDecimal(dayInt, 2);
+
+    //i really would want the date to be dynamic based on what's in the config, but the most efficient thing to do right now is simply two if statements covering the 2 common conditions.
+    if (DATE_FORMAT == "mm/dd/yyyy") {
```
uvwxy, line 40: can we change this that it still works when `DATE_FORMAT` is not defined?

I've seen a couple of users stumbling across not updating their `config.h` respectively.

If we have a default case for `TIME_FORMAT` and `DATE_FORMAT` not being defined in addition would be optimal.

Thanks you :)

uvwxy (changes requested):

src/apps/main/watchface_digital.cpp
rawsignal: Alright, both TIME_FORMAT and DATE_FORMAT should be protected in current uses against being undefined

rawsignal:
Since I have 2 pr's open, in some cases against the same files, there will be merge conflicts when one of the PR's gets approved and merged. I'll keep an eye out for when the first one gets merged and take care of getting the changes from that merged into the unmerged branch

uvwxy:
@rawsignal  you underestimated the power of git ;)

Issue #67, opened 2021-05-06
Add improved button handling methods


uvwxy:
- btnHasGoneDown(BTN_X)
	- btnHasGoneUp(BTN_X)
	- btnIsDownSince(BTN_X)
	- btnIsUpSince(BTN_X)
	- btnIsLongPress(BTN_X)

uvwxy:
@xasz

uvwxy:
I merged your pull request. thank you !

Issue #55, opened 2021-05-02
General Conception of Apps/Menus and usage design.


xasz:
As i follewed and saw on Github there are many people including me testing around with menues and diferrent types of apps/watches and app switching.

I would love to setup a "design" language for all of us and an technical one as well.

### Current Concept:
We have Apps. Everything is a App and can to whatever it wants to.
Btn1 is the app-switch by long press and no app can change that.

### My Idea:
I would split up everything in 3 parts:

**Part 1: Launchers**
A provides the functinality to present Aps in Faces, in whatever way it chooses to. The Launcher itself can use all 3 availible Buttons to switch to Menues/Faces/Apps. A Launcher although is used to show overlays or "UI" specific things.

**Part2: Faces**
A Face can be a Watchface itself or any infomational like stepcounter/calendar/stopwatch whatever, but it is limited to exctly 1 button press. and thats chooses not by the face, but instead by the launcher. The Launcher sends this trigger to the face.
The abstract face class would also provde easy access for get Current Date/Time/Information that nobody needs to right these things over and over themself in each face.

**Part3: Apps**
A App is started by launcher and has full control over everything but also have to be able to have a "back to launcher button".

### Full Example:

**Launcher: xaszLauncher**
My Launcher decides that it toggels between Faces on Btn1 and Bt3. 
Btn2 is the trigger for Faces. Every Other Launcher could use other buttons or a combinations.

**My Face1: Watchface**
My custom or your custom Watchface what i like to have on my wrist. The Trigger Action could Start my Faverite configred App or go to the Main Menu

**My Face1: Fitness**
Shows my current Stepcounter and the triggerpress could start/stop and workout timer.

**App: Settings**
There should be a general Settings App for the launcher itself and "Display/Settings".

**App: RandomOne**
At some point i want to make a game :)

The Essential part of this is the convention of who has control over what.
If we setup a construct of abstract classes for Launcher (I am on it) and Face and App and who has the button control we could easy swap out one for  #another.
There could be many Faces which are compatible with every launcher and if you don't like the launcher but your faces, just use another one.

I have thought about the concept alot and i think we are limited with 3 buttons but i think to have well working launchers/faces/app in the community we need this kind of convention and technial abstract classes everyone can then use.

Like
osw_launcher
osw_face
osw_app

If you like the idea or the concept i would be happy to code the ground construct and pull it again.
I think it would be very nice for all of us and makes swapping apps/faces much more easier in the future.

[...]

PR #68, opened 2021-05-06
Button handling

xasz:
Added Methods for Buttons like Game Engine do:

` // Buttons (Engine-Style)
  bool btnHasGoneDown(uint8_t btn);
  bool btnHasGoneUp(uint8_t btn);
  bool btnIsPhysicalDown(uint8_t btn);
  bool btnIsLongPress(uint8_t btn);`

U can now easy acess events in your app-loop like:

`if(hal->btnHasGoneDown(2)){
  startMyAction();
}`

Added a small app to test the events.

Issue: #67

### DIFF
--- a/include/osw_hal.h
+++ b/include/osw_hal.h
@@ -16,6 +16,12 @@ using std::string;
 
 #define BTN_CLICK_TIMEOUT 333
 
+enum PhysicalInput {
+  BUTTON_1 = 0,
+  BUTTON_2 = 1,
+  BUTTON_3 = 2
+};
+
 class OswHal {
  public:
   // Constructor
@@ -46,6 +52,12 @@ class OswHal {
   bool btnHasGoneUp(uint8_t btn);
   bool btnIsPhysicalDown(uint8_t btn);
   bool btnIsLongPress(uint8_t btn);
+  void suppressButtonUntilUp(uint8_t btn);
+
+  
+
+
+
 
 
   // Display
@@ -154,7 +166,7 @@ class OswHal {
   bool btnGoneUp[3];  
   bool btnSuppressUntilUpAgain[3];  
   bool btnGoneDown[3];   
-  long btnIsDownSince[3];  
+  unsigned long btnIsDownSince[3];  
   bool btnLongPress[3];  
   long _lastTap = 0;
   long _lastDoubleTap = 0;
--- a/include/osw_hal.h
+++ b/include/osw_hal.h
@@ -40,6 +40,13 @@ class OswHal {
   void clearBtn1(void);
   void clearBtn2(void);
   void clearBtn3(void);
+  
+  // Buttons (Engine-Style)
+  bool btnHasGoneDown(uint8_t btn);
+  bool btnHasGoneUp(uint8_t btn);
+  bool btnIsPhysicalDown(uint8_t btn);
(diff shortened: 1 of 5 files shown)


include/osw_hal.h
```diff
@@ -142,6 +149,13 @@ class OswHal {
   long _btn1UpSince = 0;
   long _btn2UpSince = 0;
   long _btn3UpSince = 0;
+  bool btnLastState[3];
+  bool btnIsDown[3];
+  bool btnGoneUp[3];  
+  bool btnSuppressUntilUpAgain[3];  
+  bool btnGoneDown[3];   
+  long btnIsDownSince[3];
```
SeanReg, line 157: Should be `unsigned long`

src/hal/buttons.cpp
```diff
@@ -46,12 +62,53 @@ void OswHal::checkButtons(void) {
...
+  btnIsDown[2] = !digitalRead(BTN_3) == LOW;
+
+  for(uint8_t i = 0  ; i < 3 ; i++){
+    btnGoneUp[i] = btnLastState[i] == true && btnIsDown[i] == false;
+    btnGoneDown[i] = btnLastState[i] == false && btnIsDown[i] == true;
+    
+    if(btnGoneDown[i]){
+      btnIsDownSince[i] = millis();  
+    }
+    btnLongPress[i] = btnIsDownSince[i] < millis() - DEFAULTLAUNCHER_LONG_PRESS && btnIsDown[i] == true;
```
SeanReg, line 78: `millis() > btnIsDownSince[i] + DEFAULTLAUNCHER_LONG_PRESS` would protect against overflow issues

SeanReg:
Some sort of input map would help here. Such as:

`
enum Input {
   INPUT_MIN

    BUTTON_1,
    BUTTON_2,
    BUTTON_3,

   INPUT_MAX
}

//Map inputs to pins
InputMap = {
   BTN_1,
   BTN_2,
   BTN_3
}
`

You could then setup the pin modes and iterate over the buttons using these. Something like:

`
for (unsigned int in = INPUT_MIN + 1; in != INPUT_MAX; ++in) {
  unsigned int idx= in - 1;
  Input inputId = static_cast<Input>(in);

  digitalRead(InputMap[idx])
}
`

This would remove a lot of the harcoded numbers and allow future buttons or inputs to more easily be added in the future if ever needed.

src/hal/buttons.cpp
xasz: Yeah we can chance that, when its switched to unisgned :) Will chance that.

xasz:
@SeanReg Thanks for the review and the idea is great.
I thought about some enum aswell but i was not sure. I had a talk with Paul yesterday about #55 and  he wants to keep stuff as simple as possible and i am fine with that. Thats why i did not use any castings and make it simple to understand.

What could be a solution/compromiss would be some simple defines for the buttons which then can be used in the code like:

`
  #define BUTTON_1 = 0
  #define BUTTON_2 = 1
  #define BUTTON_3 = 2
`

@uvwxy What would u like to have implemented ?

uvwxy:
The enum is fine 👍

xasz:
Perfect.

xasz:
i added a new commit which hope does make happy everyone for now :)

PR #69, opened 2021-05-07
Add digital watchface to main app list

jlevy-dev:
-I added the digital watch-face as the second app
-Set the second app (digital watch-face) to have auto sleep 
-Raised auto sleep to 10 seconds as the 5 seconds was way to short

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -39,7 +39,7 @@ OswHal *hal = new OswHal();
 #if defined(GPS_EDITION)
 #define NUM_APPS 5
 #else
-#define NUM_APPS 4
+#define NUM_APPS 5
 #endif
 RTC_DATA_ATTR uint8_t appPtr = 0;
 OswApp *mainApps[] = {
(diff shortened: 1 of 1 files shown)


jlevy-dev:
Do not push yet, on device it is acting funny and only working on the digital and not analog

src/main.cpp
```diff
@@ -160,7 +161,7 @@ void loop() {
   }
 
   // auto sleep on first screen
-  if (appPtr == 0 && (millis() - appOnScreenSince) > 5000) {
+  if (appPtr == (0 || 1) && (millis() - appOnScreenSince) > 10000) {
```
SeanReg, line 164: (0 || 1) evaluates to true (1). You want something like:

`if ((appPtr == 0 || appPtr == 1) && (millis() - appOnScreenSince) > 10000)`

jlevy-dev:
Thanks @SeanReg that worked and tested on device, putting both watch-faces to sleep.

jlevy-dev:
awesome thanks!

Issue #73, opened 2021-05-09
Set time manually


mat942:
Given some hardware boards have wifi issues, an option to set the time manually would make them usable.

richtepa:
If no one is working on it, I'd love to start doing that part tonight.

uvwxy:
I've merged @richtepa's pull request (Thank you!)

I will improve it a little bit (structuring and allow decrements as well)

PR #80, opened 2021-05-11
added possibility to change the time manually

richtepa:
#73 

There are at least two things to think about:

- Does the location (time_from_web.cpp) of this feature makes sense? Should it be seperated?
- Is the UI/UX okay? Feedback and ideas are welcome.


uvwxy:
I will fix the broken build once I've merged it

PR #82, opened 2021-05-12
Twist Wrist to Wake

rasppliances:
Added twist wrist to wake with changes to sensors.cpp and power.cpp

### DIFF
--- a/src/hal/power.cpp
+++ b/src/hal/power.cpp
@@ -103,6 +103,8 @@ void OswHal::deepSleep() {
   // rtc_gpio_isolate(GPIO_NUM_35);
   esp_sleep_enable_ext0_wakeup(GPIO_NUM_0 /* BTN_0 */, LOW);
   esp_sleep_enable_ext0_wakeup(GPIO_NUM_35 /* BMA_INT_2 / TAP */, HIGH);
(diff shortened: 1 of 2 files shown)


uvwxy:
I will take a look at this today, but I'll merge it into develop first.

xasz:
@rasppliances Nice work
Seems to work fine. I cannot tell if this is the right way todo, because i do not know about the hardware much, but it works.
I think there need to be some playing aroung with the threshold because i think it feels very low.

What i did't not understand is this in the sensors.cpp?
Is this a copy paste error?

`rslt = bma400_get_sensor_conf(accel_setting, 3, &bma);
  bma400_check_rslt("bma400_get_sensor_conf", rslt);
  rslt = bma400_get_sensor_conf(accel_setting, 3, &bma);
  bma400_check_rslt("bma400_get_sensor_conf", rslt);`

rasppliances:
@xasz Thanks!

The threshold certainly is flexible.  In my testing it was OK, but it was an arbitrary number.  I also didn't test it on my right wrist, so that might make a difference.

That code is a copy paste error.  I don't know why I didn't see that.

uvwxy:
I cherry picked your commit: 9987773

PR #83, opened 2021-05-13
Prevent Deepsleep while buttons are pressed

xasz:
If the deepsleep gets triggered while a button is pressed, we reset the screentime for the app.
This should avoid deepsleep while we do stuff on the watch :)


xasz:
No idea why this fails. only changed one file. Anyone a idea ?

PR #90, opened 2021-05-16
Sprite system for graphics handling

SeanReg:
- Sprite and ShapeSprite base class
- Implementations for rectangle, circle, triangle, text
- Set/get position, alignment, draw layer, size, etc
- Renderer which draws all Sprites with a single call
- Sprites automatically add/remove themselves too/from renderer
- Renderer can "clearAll" Sprites and optionally dealloc them all

General flow:

App Setup()

Instance and setup desired sprites

App Loop()

Control, move, etc sprites
Call Renderer.draw() to display the sprites

App Stop()
Call Renderer.clearAll(true) to release all the Sprites.

The might also work well for System Overlays. Just would need to make some changes so clearAll doesn't clear the Overlays (Maybe set aside a special layer for overlays?)

### DIFF
--- a/include/osw_hal.h
+++ b/include/osw_hal.h
@@ -23,6 +23,9 @@ using std::string;
 // enum for user space button handling
 enum Button { BUTTON_1 = 0, BUTTON_2 = 1, BUTTON_3 = 2 };
 
+#define SCREEN_WIDTH 240
+#define SCREEN_HEIGHT 240 
+
 class OswHal {
  public:
   // Constructor
--- a/src/sprites/circle_sprite.cpp
+++ b/src/sprites/circle_sprite.cpp
@@ -1,0 +1,44 @@
+#include <sprites/circle_sprite.h>
+
+void CircleSprite::setSize(uint16_t x, uint16_t y) {
+    //Circle has equal sizes
+    Sprite::setSize(x, x);
+}
+
+uint16_t CircleSprite::getRadius() {
+    return Sprite::getSizeX(); 
+}
+
+void CircleSprite::setRadius(uint16_t r) {
+    setSize(r * 2, r * 2);
+}
+
+
+int16_t CircleSprite::getOffsetX() {
+    if (getAlignment() & SpriteAlignment::LEFT) {
+        return -(getSizeX() / 2);
+    } else if (getAlignment() & SpriteAlignment::RIGHT) {
+        return (getSizeX() / 2);
+    }
+
+    return 0;
+}
+
+int16_t CircleSprite::getOffsetY() {
+    if (getAlignment() & SpriteAlignment::TOP) {
+        return -(getSizeY() / 2);
+    } else if (getAlignment() & SpriteAlignment::BOTTOM) {
+        return getSizeY() / 2;
+    }
+
+    return 0;
+}
+
+void CircleSprite::draw(OswHal *hal) {
+    int16_t posX = getX() - getOffsetX(), posY = getY() - getOffsetY();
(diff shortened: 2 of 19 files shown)


include/osw_hal.h
```diff
@@ -23,6 +23,9 @@ using std::string;
 // enum for user space button handling
 enum Button { BUTTON_1 = 0, BUTTON_2 = 1, BUTTON_3 = 2 };
 
+#define SCREEN_WIDTH 240
```
max9901, line 26: This is already defined in config.h 
```
#define DISP_W 240
#define DISP_H 240
```
consider using:
hal->gfx()->getHeight() 
hal->gfx()->getWidth()

src/sprites/circle_sprite.cpp
```diff
@@ -0,0 +1,44 @@
...
+        return getSizeY() / 2;
+    }
+
+    return 0;
+}
+
+void CircleSprite::draw(OswHal *hal) {
+    int16_t posX = getX() - getOffsetX(), posY = getY() - getOffsetY();
+    if (getFilled()) {
+        hal->getCanvas()->fillCircle(posX, posY, getSizeX() / 2, getColor());
```
uvwxy, line 40: The recommended way is to use `hal->gfx()` instead of `hal->getCanvas()`

src/sprites/triangle_sprite.cpp
```diff
@@ -0,0 +1,65 @@
...
+    int16_t y0 = getY() - getSizeY() / 2 - getOffsetY();
+
+    int16_t x1 = getX() - getSizeX() / 2 - getOffsetX();
+    int16_t y1 =  getY() + getSizeY() / 2 - getOffsetY();
+
+    int16_t x2 = getX() + getSizeX() / 2 - getOffsetX();
+    int16_t y2 = getY() + getSizeY() / 2 - getOffsetY();
+
+    if (getFilled()) {
+        hal->getCanvas()->fillTriangle(x0, y0, x1, y1, x2, y2, getColor());
```
uvwxy, line 61: `draw/fillTriangle` unfortunately don't exist yet in `gfx_2d.h`.

At some point there _might_ be a different display involved, that is likely not supported by `ArduinoGFX`. In this case we only need to flush the 2d buffer.

src/sprites/circle_sprite.cpp
SeanReg: Little confused on the purpose of gfx(). Canvas is already buffered and it looks like gfx is actually allocating a second frame buffer?

uvwxy: Gfx(gfx_2d) is the buffer that canvas writes to. The desktop code only uses the drawfunctions of gfx_2d.

SeanReg: A little confused why we the have gfx_2d. For the desktop emulation I think we should still be using `ArduinoGraphics2DCanvas` but for the `Arduino_G` passing it an adapter which writes to the framebuffer of a desktop application?

uvwxy: sorry for the late reply...

yes, this way we can cross compile the code on desktop and do experiments using the same renderer using an SDL window (src: [matrix.cpp](https://github.com/Open-Smartwatch/lib-open-smartwatch/blob/master/examples/matrix/main.cpp)):

demo

simonmicro:
I'm considering closing this issue, as it is seriously out-of-sync with the current `master` of the OS and has not received any updates since ages...

Issue #54, opened 2021-05-01
Configuration Abstraction + Configuration UI


Simonmicro:
This should store minimal key - value pairs using `Preferences.h`.

Store some things like:
* Last open app (so when we crash again, the user will not notice too much)
* Last connected BLE device (to reconnect on boot)
* App data? Stopwatch stuff?
* Boot count & flash write counts
* Storage version (if version increase, wipe storage on boot & populate with defaults)

All keys are handled by a enum, available inside the respective HAL. This eliminates any name collisions & minimizes the nvs use (as we only need to store one char for the first 26*2+10 keys)...

But we should somehow minimize the write cycles on this part (only static data?), otherwise we literally kill the memory over time -> therefore every write operation must explicitly be enabled...

uvwxy:
Step 1: Configuration Abstraction
- wrap Preferences.h (ESP32 lib) 
- define which keys exist (typing versus config.json) (in a single place, e.g. name/type/default inside a header)
- examples how to use it

Step 2a: Configuration UI
- implement functions to parse/generate data.json (POST/GET)
- implement functions to generate config.json (GET)

Step 2b: Configuration UI
- include bundle.js/css/index.html as header files
- expand wifi connect / time sync application to start the config server and show the IP


Step 3: Optional / Future
- if the wifi is not connectable -> create a wifi hotspot
- generate wifi password dynamically (once ;) ) and display on screen
- HTTPS
- password protect the configuration
- include contributors page

Simonmicro:
Okay, I finished step 01 - see here: https://github.com/Simonmicro/open-smartwatch-os/tree/feature/nvsConfig
The `config.json` should be able to use the `OSW_CONFIG_*` defines (inside an other one) to be "generated" during compile-time later on. I also created an other branch `debug/heapy` to further investigate the system crashes - for now no results.

Simonmicro:
And here are the `#define`s for the `config.json` to support the expansion of the config keys (just a quick demo program):
```c++
#include <iostream>
#include <string>

//These are the config key definitions from the osw_config.h
#define OSW_CONFIG_WTF_COLOR_R "a", "s", 179
#define OSW_CONFIG_WTF_COLOR_G "b", "s", 107
#define OSW_CONFIG_WTF_COLOR_B "c", "s", 0

//This defines a section of the json (the # allows insertion of params, even between strings)
[...]

PR #91, opened 2021-05-17
Feature/config caching

Simonmicro:
Forced config key handling back to the OswConfig part & introduced config key value caching for all primitive datatypes.


Simonmicro:
This closes #54 !

PR #96, opened 2021-05-20
Feature/Services2.0

Simonmicro:
- Refactored @SeanReg service system
- Added example service
- Added memory monitor service
- Added GPS service (untested!)
- Cleanup


src/services/OswServiceTasks.cpp
```diff
@@ -0,0 +1,45 @@
...
+#ifdef GPS_EDITION
+OswServiceTaskGPS gps;
+#endif
+//OswServiceTaskExample example;
+#ifdef DEBUG
+OswServiceTaskMemMonitor memory;
+#endif
+}  // namespace OswServiceAllTasks
+
+const unsigned char oswServiceTasksCount = 0 + SERVICE_BLE_COMPANION + SERVICE_GPS + SERVICE_MEMMONITOR;
```
uvwxy, line 33: I was thinking about how we would do this one 👍 
too simple :)

src/services/OswServiceTaskMemMonitor.cpp
```diff
@@ -0,0 +1,52 @@
+#include "./services/OswServiceTaskMemMonitor.h"
+
+#include "osw_hal.h"
+#include "services/OswServiceManager.h"
+
+void OswServiceTaskMemMonitor::setup(OswHal* hal) { OswServiceTask::setup(hal); }
+
+void OswServiceTaskMemMonitor::loop(OswHal* hal) {
+  unsigned core0 = uxTaskGetStackHighWaterMark(nullptr);
```
uvwxy, line 9: I think I could benefit from some comments to understand this

uvwxy (approved):

src/services/OswServiceTasks.cpp
Simonmicro: :grin: - is this comment done?

src/services/OswServiceTaskMemMonitor.cpp
Simonmicro: OK - comes soon :tm:

PR #97, opened 2021-05-20
Fix time format config

rawsignal:
replacing config entry for time format


rawsignal:
some strangeness as a result of rebase. closing

Issue #99, opened 2021-05-21
Setting of wake sensibility


mat942:
Latest releases leave the OSW most of the time on which increases power consumption. Manual setting of this setting to each user's taste would be a nice addition

uvwxy:
I added settings to disable/enable this mode and cherry-picked the changes from #100 to allow threshold editing

uvwxy:
Commit with a082bbd

PR #100, opened 2021-05-22
Feature/raise to wake toggle

rawsignal:
I think this might be what we need to make the sensitivity adjustable, but I haven't been able to experiment/test with my watch being out of commission for a bit. In the mean time, maybe someone who knows more about the hardware etc would be able to identify if this is on the right track or completely wrong. this is related to https://github.com/Open-Smartwatch/open-smartwatch-os/issues/99


uvwxy:
I cherry-picked your change 0d9b47d

and added a toggle a082bbd 

Thank you!

PR #101, opened 2021-05-22
Verifies if raise to wake feature is turned on

airtonzanon:
For the raise to wake feature be working 100% we need to also verify it
in the power file.


src/hal/power.cpp
```diff
@@ -104,7 +104,9 @@ void OswHal::deepSleep() {
   // rtc_gpio_isolate(GPIO_NUM_35);
   // esp_sleep_enable_ext0_wakeup(GPIO_NUM_0 /* BTN_0 */, LOW);
   // esp_sleep_enable_ext0_wakeup(GPIO_NUM_35 /* BMA_INT_2 / TAP */, HIGH); // step interrupts (currently)
-  esp_sleep_enable_ext0_wakeup(GPIO_NUM_34 /* BMA_INT_1 */, HIGH);  // tilt to wake and tap interrupts
+  if (OswConfigAllKeys::settingRaiseToWakeEnabled.get()) {
+    esp_sleep_enable_ext0_wakeup(GPIO_NUM_34 /* BMA_INT_1 */, HIGH);  // tilt to wake and tap interrupts
+  }
```
uvwxy, line 109: tap interrupts are also on BMA_INT_1

PR #102, opened 2021-05-22
Allowing on-device brightness updates

rawsignal:
Adding new config value - minimum brightness. This is to prevent the user from setting the brightness to 0 on accident and forgetting about it. The minimum can be changed to 0 if the user would like to be able to set the brightness to 0. This change persists the brightness the user selects on-device. Might be nice to introduce some kind of helper method or something in OswConfig that handles enableWrite and disableWrite for reducing repetitive code.

Tested working fine thus far.

### DIFF
--- a/include/config.h.example
+++ b/include/config.h.example
@@ -21,7 +21,7 @@
 #define DEVICE_NAME "yourDeviceName"
 #define BLE_DEVICE_NAME DEVICE_NAME
 
-
+#define DISPLAY_MIN_BRIGHTNESS 10
 
 // The following settings are configureable later on using the web ui, you can still set the defaults here.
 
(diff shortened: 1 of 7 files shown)


Simonmicro:
Hi,
good idea, but is there really a need that the user can change this value using the config UI? I _personally_ think it would be better to compile this in, rather than using the config storage. This would still prevent the user from dimming the screen too far, but would also allow him to modify it using the `config.h` when he _really_ wants to change this.
Thanks,
~Simonmicro

rawsignal:
> Hi,
> good idea, but is there really a need that the user can change this value using the config UI? I _personally_ think it would be better to compile this in, rather than using the config storage. This would still prevent the user from dimming the screen too far, but would also allow him to modify it using the `config.h` when he _really_ wants to change this.
> Thanks,
> ~Simonmicro

That's probably fair. I can switch it up a bit.

Simonmicro:
Perfect! :ok_hand:

Issue #107, opened 2021-05-29
Combine config & time-config


richtepa:
According to the [new App-Switcher](https://discord.com/channels/673521025593638932/748973564610215947/839938087571161088) we could combine the config app (config_mgmt) and time-config app (time_config) into a single settings app. This way it only uses one tile in the AppSwitcher.

Is this change desired? If so, I would like to do that.

PR #108, opened 2021-05-29
Combined Settings App

richtepa:
I also added the possibility to temporarily disable the inner-App switcher (short press).
This is needed to use button 1 while setting the time manually.

closes: #107


richtepa:
I also added a Pagination Indicator

PR #135, opened 2021-07-22
Config.h.example - remove SLEEP_BUTTONWAKE 

mpegg4:
Remove SLEEP_BUTTONWAKE from example config file in favor of develop PR that introduced default value for "WAKE_FROM_BTN1" (default false):

https://github.com/Open-Smartwatch/open-smartwatch-os/commit/37d7d51ac3ae36a702939656cdf79f107e6adcc6

Keeping SLEEP_BUTTONWAKE in example config will be incorrect.


mpegg4:
thank you @Simonmicro

Issue #65, opened 2021-05-06
Step counter does not reset at the end of the day


rasppliances:
The step counter never resets.  It should reset every day at midnight.

Another thing about the step counter...

The target is hard coded at 360.  It would be good if we could put the default value of something more realistic like 10,000 or 5,000 in config.h.  Additionally, if we store the target in the flash memory using Preferences library, we could also write an app to adjust the goal right on the watch.

max9901:
Would it make sense to use the <Preferences.h> lib with a step-count namespace to store step values. And then write a OSWservice to deal with updating and resetting step values ?

uvwxy:
Yes 💯

max9901:
We do have a problem with reading out the time from the RTC, i was checking it tonight but we get errors (aka 0 values) when 2 cores are using the i2c bus at the same time. Should we maybe make a general system background process that actually does the time updates etc instead of giving the user full control of querying the I2c line ?

uvwxy:
Also yes :)

max9901:
[WIP]~
https://github.com/max9901/open-smartwatch-os
If time permits let me know what u think of the approach :D

max9901:
So i Have some issues with I2C on core 0. The counter reset idea is implemented where we save the last 7 days in nvs-mem.
I only want to move all I2c Sensor updates to a service on core0. However, this give very in-stable readings somehow.

uvwxy:
Recent develop branch has a step history of 7 days, and resets the first time the watch starts on a new weekday, resetting the counter and carrying over the steps to the previous day.

PR #136, opened 2021-07-27
PR for #65

max9901:
Hey guys, 

Here the PR to the development branch because i am not fully satisfied yet because there is no validation right now on the step-counting reset. 

However i think its a good first prototype. 

things i changed:
- add a step-count reset at midnight locale time
- store the last 7 days of steps in nvm memory and make them accessible trough the HAL object
- Make fetching the time for the RTC an stand alone process and add error checking here
- some cleanup to force even faster startups and some personal flavors probably. 

i would love to discuss the road forward if anyone is up-to it. 

cheers,
Max


src/main.cpp
```diff
@@ -115,19 +143,18 @@ void setup() {
...
-  static long lastFlush = 0;
-  static boolean delayedAppInit = true;
-
-  // check possible interaction with ULP program
-  loop_ulp();
 
+  mainAppSwitcher->loop(hal);
   hal->handleWakeupFromLightSleep();
   hal->checkButtons();
+  hal->fetchRtcTime();         //TODO - doesn't need every cycle update.
```
Simonmicro, line 150: Is there any chance to finalize this before accepting this PR?

src/services/OswServiceTaskCore.cpp
```diff
@@ -0,0 +1,112 @@
+#include <time.h>
```
Simonmicro, line 1: Could your rename this task to `Steps`? As the task currently _only_ handles the step counting with the history and the name `core` is ambiguous.

include/osw_hal.h
```diff
@@ -6,6 +6,7 @@
 #include <gfx_2d_print.h>
 
 #include <string>
```
Simonmicro, line 8: Could you try to use the Ardunio String library? It can be (always) used via `String` and does not require you to include the `std::string` lib.

src/apps/watchfaces/watchfacemax.cpp
```diff
@@ -0,0 +1,111 @@
+#include "./apps/watchfaces/watchfacemax.h"
+#include "./fonts/Roboto_Condensed_Light_21.h"
+#include "./fonts/Roboto_Condensed_Light_75.h"
+
+#include <string>
```
Simonmicro, line 5: Same here.

Simonmicro (changes requested):
Hey, nice work - here some requests, which must not be completely implemented, but would be very nice to have.

Simonmicro:
Additionally could you try to rebase to the current `develop` branch?

Thx!

uvwxy:
Nice work @max9901  , and thanks for the review @Simonmicro

src/main.cpp
max9901: Yeah that's easy but it's a trade-off we can just place it in the 30fps limiter of the screenflush. I wasn't sure which trade-off i wanted to make here. What would u suggest ?

src/services/OswServiceTaskCore.cpp
max9901: Yeah my end goal is to actually perform all core functions here like:

```
  hal->handleWakeupFromLightSleep();
  hal->checkButtons();
  hal->fetchRtcTime();
 ```
What are your toughts ?

File: include/osw_hal.h (hunk 3 above)
Comment (max9901) [reply include/osw_hal.h:8]: sure this wasn't me btw this is already in the development branch.

**Comment by max9901** [comment]
I already am in sync with the current `develop` branch right ? or am i missing something ?

File: include/osw_hal.h (hunk 3 above)
Comment (Simonmicro) [reply include/osw_hal.h:8]: Okay that's on @rawsignal !

**Comment by Simonmicro** [comment]
You forked off a fork an therefore have an other history - which is fine for merging, but your history (and changes) would by much cleaner, when you only commit _your_ changes (meaning only after a60e180 ). That's something for the future, but also nice to correct on this one.

File: src/services/OswServiceTaskCore.cpp (hunk 2 above)
Comment (Simonmicro) [reply src/services/OswServiceTaskCore.cpp:1]: >  hal->handleWakeupFromLightSleep();

YES

>  hal->fetchRtcTime();

Yes

>  hal->checkButtons();

Maybe? This is user interaction and therefore time critical - there is grantee that we won't slow down the services further down the road. You know what? I'll open an issue to rewrite the manager to respect time-critical services.

Comment (Simonmicro) [reply src/services/OswServiceTaskCore.cpp:1]: But for now, it does not has those - right? Therefore for now, it is only the step service. Also, why would you try to merge all those functions into one service task - wouldn't many small be better? Stack size... Yes... Okay, point made for now...

File: src/services/OswServiceTasks.cpp [hunk 5]
```diff
@@ -11,23 +12,25 @@
 #include "services/OswServiceTaskWebserver.h"
 
 namespace OswServiceAllTasks {
-// OswServiceTaskExample example;
```
Simonmicro, line 14: Hey, this should stay - for anyone wanting to fiddle around with the example task (and learn how to use our services)!

src/main.cpp
Simonmicro: > This is btw the reason also why im have a PR to the develop branch :D

In that case, you could mark your PR as a draft. :grin: 

>  I wasn't sure which trade-off i wanted to make here.

Sry, but what are the tradeoffs? Fetch every cycle or... Every frame? The latter sounds better.

src/services/OswServiceTasks.cpp
max9901: it's a few lines lower but it's still there.

src/main.cpp
max9901: if we include it in the display refresh things could break we certain apps don't update the screenbuffer as much. Simply because we fill the screen buffer first than than cal the display refresh function. We would be always a frame to late. and in the case of only 1 screenrefresh every 3 seconds for example this would break stuf.

Simonmicro: Ahh, got it. Okay - then it is fine there. Additionally the system should have always the newest time cached anyways.

src/services/OswServiceTasks.cpp
Simonmicro: Whopps, missed that on my second review! :facepalm:

max9901:
can we agree to merge this with a squash merge? so i don't need to rebase?

Simonmicro (approved):

Simonmicro:
As this PR includes multiple (deep) changes (potentially breaking ones), I will not merge this on my own.

@uvwxy @SeanReg could you try to daily drive this for a while? I'll do this myself as soon it is in sync with the newest `develop` - @max9901 could you merge the current `develop` (and then ping me)? @uvwxy had fixed some stuff on there.

Some things to look out for:
* startup times really better (DeepS/LightS)?
* time still in sync (from wifi, manual, stopwatch)
* step-counter on zero when you pick it up
* something, something watchface?

uvwxy:
@max9901 I would like to go over your changes and cherry pick them onto master. do you have some time this week to look over this together via discord/screenshare?

max9901:
yes we can arrange that, can u send me a private message on discord ? so we can pick a timeslot. I can do tomorrow before 1500 the whole day and the rest of the week only in the evenings.

uvwxy:
I think your user name differs on discord. You need to help me out a bit here 😅

max9901:
maxe#2046   and im Max Eu in your channel.

uvwxy:
Thank you a ton for your work!

I looked at your commit. squashing + merge / or merge / or rebase did not result in a mergeable result, I carried over your idea with db77c28 but running everything in the first core.

@Simonmicro rule of thumb: no i2c from background tasks to avoid collisions for now. all i2c reads should happen at the top of the main loop, or from an app. if we need data from an external device in the background we need to add semaphores.

PR #167, opened 2021-10-29
added example for using fonts

iksi4prs:
added example for using fonts - both ready (GFXfont downloaded from adafruit) and custom (TTF converted to GFXfont)


src/main.cpp
```diff
@@ -146,6 +147,8 @@ void loop() {
     // tools
     mainAppSwitcher->registerApp(new OswAppStopWatch());
     mainAppSwitcher->registerApp(new OswAppWaterLevel());
+    // examples
+	mainAppSwitcher->registerApp(new OswAppFontsExample());
```
simonmicro, line 151: My only request is to not include this developer demo into the main apps - we already have multiple apps and I doubt it is an good idea to add an other app to them, as our switching is already bulky enough.

iksi4prs: Do you want me to change the commit, or edit it yourself ? (can just comment line 151, so if someone is interested in checking the example it's easy for him to find and build)

simonmicro: Please edit your commit or push a new one. Thx!

iksi4prs: done.

iksi4prs: @simonmicro , I see there was a "build_linux" error, but not details. I build on Windows. Anyway to see the error somewhere ? 10x.

simonmicro: Check out the build logs here: https://github.com/Open-Smartwatch/open-smartwatch-os/runs/4137185127?check_suite_focus=true

Try to run the used script yourself - if it still does not work you may have broken something... Or the CI is bad again - argh!

simonmicro: WAIT. Your PR is so old, that you are not using our current CI script, which had this behavior fixed! Could you rebase your changes on the current `develop` and force-push it? This should fix it and prevent any incompatibility with our current code-base...

iksi4prs: @simonmicro , I spent 1 hour trying to fetch the upstream and syncing the submodules, but eventually after rebase there were always some build errors. I think the faster solution is just to create a new branch and PR. Just for confirmation - you suggest I branch and PR from/to "develop" instead of "master" ? 10x

simonmicro: Okay, I'll do it instead. I hope tomorrow to have some time to fix it - I'll close this PR whenever I'm done. Git can sometimes be... Funny 🙃

iksi4prs: @simonmicro , 10x. But for future PRs, should I use "develop" instead of "master" ? 10x

simonmicro: Definitively.

iksi4prs: @simonmicro , ok , 10x for your help.

PR #178, opened 2021-12-01
Update : font patch

RuffaloLavoisier:
The steps to fix the font patch, modify it for smooth work.


simonmicro:
This won't build. Did you tested that?

RuffaloLavoisier:
@simonmicro Not yet, I just added it to use it as a test or not when the patch is completed later.

RuffaloLavoisier:
@simonmicro Check - I modified the font file on a large scale. There is no problem with the compilation process. I will check the implemented display problem later while developing the UI.

simonmicro:
> I modified the font file on a large scale. There is no problem with the compilation process.

:+1: 

> while developing the UI

What are you doing? I thought you tried to fix the fonts - are you now also adding other stuff? In case this is getting bigger, please consider to split up your contributions.

RuffaloLavoisier:
@simonmicro Then, i will finish modifying the font file. i will reopen it when there is another commit later. I'd appreciate it if you could review the merge.

PR #179, opened 2021-12-01
Update : debug

RuffaloLavoisier:
added debugging to monitoring the current step count on watchface.cpp
And changed the name of other variables.
s -> step_num
i -> index


simonmicro:
There is no need for this logging - as it is already done inside the `statistics_steps.cpp` file.

PR #185, opened 2022-02-03
Feature/abstract platform

simonmicro:
All sensors & RTC
display etc is pending, will be added later on.


RuffaloLavoisier:
Before merging into master, I modified the library for code capacity. (https://github.com/Open-Smartwatch/open-smartwatch-os/pull/187)

Issue #197, opened 2022-03-17
Watchface-MIX


RuffaloLavoisier:
<img src="https://user-images.githubusercontent.com/66334969/158962758-50406f3e-3979-47e6-ad96-b6152f45dbcf.gif" width="60%">

I would like to add a new `Watchface-MIX`.

Let me know if there's anything I need to pay more attention to.

* Watchface-MIX is Analog Watch + Digital Watch

RuffaloLavoisier:
@simonmicro It was developed by the [libraries repository](https://github.com/Open-Smartwatch/lib-open-smartwatch). I will organize the current code and open the PR within this week.

RuffaloLavoisier:
@simonmicro 
See https://github.com/Open-Smartwatch/lib-open-smartwatch/pull/18

First, I added a virtual test development environment. Please check. I will add Watchface to the main repository today or tomorrow.

RuffaloLavoisier:
Done.

PR #199, opened 2022-03-19
Apps/watchfaces : Update Watchface-MIX

RuffaloLavoisier:
Hello World !
Finally, I will update the new watchface.

* Update Watchface-MIX
* Algorithm update due to addition of new date format
* Some update in OSW-config

See : https://github.com/Open-Smartwatch/open-smartwatch-os/issues/197

### DIFF
--- a/include/apps/watchfaces/watchface_digital.h
+++ b/include/apps/watchfaces/watchface_digital.h
@@ -8,16 +8,14 @@
 
 class OswAppWatchfaceDigital : public OswApp {
  public:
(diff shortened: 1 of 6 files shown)


RuffaloLavoisier:
There are some bugs, so I'll draft

RuffaloLavoisier:
The virtual development environment and the parameters applied by the device are different. 

I have changed the part to find the angle according to the current time.

simonmicro (approved):

PR #203, opened 2022-03-25
HAL : change delay system

RuffaloLavoisier:
The existing delay means a stop state, which is fatal in many ways to the board, so it implements the delay in a different way.

### DIFF
--- a/src/hal/buttons.cpp
+++ b/src/hal/buttons.cpp
@@ -36,13 +36,14 @@ void OswHal::setupButtons(void) {
 
 #if defined(GPS_EDITION) || defined(GPS_EDITION_ROTATED)
 
-void OswHal::vibrate(long millis) {
+void OswHal::vibrate(long delay) {
   digitalWrite(VIBRATE, HIGH);
 #ifndef NDEBUG
   Serial.print("Vibrate for: ");
-  Serial.println(millis);
+  Serial.println(delay);
 #endif
-  delay(millis);
+  long prev = millis();
+  while (millis() - prev <= delay);
   digitalWrite(VIBRATE, LOW);
 }
 #endif


simonmicro:
Could you explain in which way a stop state should be fatal? The kernel is way happier to interrupt one sleeping thread, instead of a busy one...

RuffaloLavoisier:
@simonmicro 

Looking at the basic `blink` example, as I know, delay means to pause the board to maintain its current state. 

```
void setup(){ 
  pinMode(LED_BUILTIN, OUTPUT); 
}
void loop(){
  digitalWrite(LED_BUILTIN,HIGH);
  delay(1000);
  digitalWrite(LED_BUILTIN,LOW);
  delay(1000);
}
```
In other words, it means stopping other tasks in progress on the board. 

I don't know because I haven't experienced the `delay()` used in `OSW esp32` yet, but if it's the same as the `delay()` in `Arduino board`, I think `delay()` in `OSW esp32` will also play a role in getting results through the same board stopping.

uvwxy:
`delay` is implemented as

```c++
void delay(uint32_t ms)
{
    vTaskDelay(ms / portTICK_PERIOD_MS);
}
```

`vTaskDelay` is documented as

> The amount of time, in tick periods, that the calling task should block.

Note that this blocks the *task* not the board. 

From the RTOS docs (the underlying system of EPS IDF, the underlying system of Arduino Framework we are using):

> Each executing program is a task (or thread) under control of the operating system. If an operating system can execute multiple tasks in this manner it is said to be multitasking.

See https://www.freertos.org/implementation/a00004.html

@RuffaloLavoisier I greatly appreciate your motivation for improving the OSW OS :). Just in this small case it would be a "Verschlimmbesserung".

RuffaloLavoisier:
> Just in this small case it would be a "Verschlimmbesserung".

@uvwxy 

Thanks for your comment :+1:

simonmicro:
@uvwxy Thank you - I couldn't have explained it better myself.

@RuffaloLavoisier You sould take a look how the kernel on FreeRTOS works / how the Arduino Framework creates the "root" task. After all this system is not perfect, but the usage of tasks there should give you some ideas. Also our OSW service task manager is utilizing these features too :)

RuffaloLavoisier:
@simonmicro 

Is there any other docs that I can look at this part or anything related to OSW project? anything.

PR #211, opened 2022-04-14
Apps : display goal steps ( Today / Total / Goal )

### LINKED ISSUE #207: Data information
I would like to provide information to Watchface. 

TO-DO:
---
* [x] Battery Notification (Arc or anything)
* [ ] Provides sunrise
* [x] Extend fitness app : https://github.com/Open-Smartwatch/open-smartwatch-os/pull/223
  * [x] Total distance
  * [x] Total calories consumed  
  * [x] Total steps
* [ ] Draw tracking map (With GPS)
>Bluetooth Applications to be developed in the future
---
I think we can provide this kind of information at the current step counter.

I think it is possible to extract sufficiently valuable data by storing user information(height, weight) that needs to be received by the Open-smartwatch in config.h or OSW config.

---
Related issue :

* https://github.com/Open-Smartwatch/open-smartwatch-os/issues/23

RuffaloLavoisier:
See #207 
Notification of step-counter ( Today / Total / Goal )

### DIFF
--- a/include/osw_config_keys.h
+++ b/include/osw_config_keys.h
@@ -58,6 +58,7 @@ extern OswConfigKeyShort resetDay;
 #if OSW_PLATFORM_ENVIRONMENT_ACCELEROMETER == 1
 extern OswConfigKeyInt stepsPerDay;
 extern OswConfigKeyBool stepsHistoryClear;
+extern OswConfigKeyBool settingDisplayStepsGoal;
 #endif
 extern OswConfigKeyDropDown settingDisplayDefaultWatchface;
 }  // namespace OswConfigAllKeys
--- a/src/apps/watchfaces/watchface.cpp
+++ b/src/apps/watchfaces/watchface.cpp
@@ -40,7 +40,14 @@ void OswAppWatchface::drawStepHistory(OswUI* ui, uint8_t x, uint8_t y, uint8_t w
     hal->gfx()->setTextBottomAligned();
     hal->gfx()->setTextSize(1);
     hal->gfx()->setTextCursor(DISP_W / 2, y - 1);
-    hal->gfx()->print(hal->environment->getStepsToday());
+
+    if (OswConfigAllKeys::settingDisplayStepsGoal.get()) {
+      char stepsTodayNGoal[20];
+      sprintf(stepsTodayNGoal, "%d/%d", hal->environment->getStepsToday(), max);
+      hal->gfx()->print(stepsTodayNGoal);
+    } else {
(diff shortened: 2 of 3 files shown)


simonmicro (changes requested):
Only one minor request: Could you change your new configKeyVariable to begin with `settingDisplay` to be a little bit more in line with the others?

Otherwise: Good work!

src/apps/watchfaces/watchface.cpp
```diff
@@ -40,7 +40,14 @@ void OswAppWatchface::drawStepHistory(OswUI* ui, uint8_t x, uint8_t y, uint8_t w
     hal->gfx()->setTextBottomAligned();
     hal->gfx()->setTextSize(1);
     hal->gfx()->setTextCursor(DISP_W / 2, y - 1);
-    hal->gfx()->print(hal->environment->getStepsToday());
+
+    if (OswConfigAllKeys::settingDisplay.get()) {
+      char stepsTodayNGoal[20];
```
RuffaloLavoisier, line 45: This part has been temporarily assigned an appropriate size. 
Memory will be free, but let me know if there's a better way?

include/osw_config_keys.h
```diff
@@ -58,6 +58,7 @@ extern OswConfigKeyShort resetDay;
 #if OSW_PLATFORM_ENVIRONMENT_ACCELEROMETER == 1
 extern OswConfigKeyInt stepsPerDay;
 extern OswConfigKeyBool stepsHistoryClear;
+extern OswConfigKeyBool settingDisplay;
```
simonmicro, line 61: Oh no! I've asked you to _prefix_ the chosen variable name, NOT to replace it entirely... Do you understand now? :rofl:

src/apps/watchfaces/watchface.cpp
simonmicro: Sure! As we use the ExpressIf framework (which is based (?) on Arduino, which is based on FreeRTOS) you could use the capabilities of the `String`-class to format it without using a temporary character buffer. I would recommend you to use that instead :smile_cat:

simonmicro (changes requested):

include/osw_config_keys.h
RuffaloLavoisier: @simonmicro hmm...`settingDisplayStepsGoal`?

RuffaloLavoisier:
@simonmicro 
Please check if I solved the *two issues* well.

simonmicro (approved):

PR #218, opened 2022-04-17
CI : compile all models

RuffaloLavoisier:
Perform a test build about all models supported in `platformio.ini`

### DIFF
--- a/.github/buildEverything.py
+++ b/.github/buildEverything.py
@@ -12,10 +12,11 @@ pioConfig = 'platformio.ini'
 
 # Configure editions
 editions = []
-build_models = os.popen("cat platformio.ini | grep '\[\env:.*\]\'").read()
-build_models = build_models.strip().split("\n")
-for build_model in build_models:
-    editions.append(build_model.replace("[env:", "").replace("]", ""))
+fh = open('platformio.ini', 'r')
+for line in fh.readlines():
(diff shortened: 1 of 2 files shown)


.github/buildEverything.py
```diff
@@ -11,14 +11,11 @@
...
-editions = [
-    'LIGHT_EDITION_V3_2',
-    'LIGHT_EDITION_V3_3',
-    'LIGHT_EDITION_V4_0',
-    'LIGHT_EDITION_DEV_LUA',
-    'GPS_EDITION_V3_1',
-    # GPS_EDITION_DEV_ROTATED not, as it is only for testing (right now)
-]
+editions = []
+build_models = os.popen("cat platformio.ini | grep '\[\env:.*\]\'").read()
```
simonmicro, line 15: I would strongly encourage you to use this instead:

```python
fh = open('platformio.ini', 'r')
for line in fh.readlines():
  if line.startswith('[env:'):
    print(line.strip())
fh.close()
```

...which is more platform-independent than your `cat | grep` solution...

simonmicro (changes requested):

.github/buildEverything.py
RuffaloLavoisier: I will change it for all OS users !

simonmicro (approved):

Issue #217, opened 2022-04-17
Improve the CI (FAST)


RuffaloLavoisier:
It's not a long time, but we want to speed up and increase efficiency in the development process.  
* if  end of coffee time then finish !
* [x] https://github.com/Open-Smartwatch/open-smartwatch-os/pull/219
* [x] https://github.com/Open-Smartwatch/open-smartwatch-os/pull/221

RuffaloLavoisier:
See #221

RuffaloLavoisier:
See #221

PR #219, opened 2022-04-17
CI : improve compile running time #217

RuffaloLavoisier:
We want to reduce OSW all-model compile running time.
See #217

### DIFF
--- a/.github/buildEverything.py
+++ b/.github/buildEverything.py
@@ -4,6 +4,7 @@ import os
 import shutil
 import logging
 import subprocess
+from datetime import datetime
 logging.basicConfig(level=logging.INFO)
 
 # Configs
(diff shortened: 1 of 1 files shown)


uvwxy:
Does this improve buildtime on github? Or just locally? My expectation would be this only improves local builds. Would be positively surprise if this also would speed CI builds.

Lgtm though :)

RuffaloLavoisier:
@uvwxy 
Expectations are that CI build speeds will be improved.

simonmicro:
I think this will also improve build times on GitHub... BUT:

- As far as I recall the `pio` command already uses more than one thread... Right? What is the default of `-j`? At least on my PC it is already blazingly fast...
- To retrieve the cpu thread count, please take a look [here](https://stackoverflow.com/a/68285053/11664229) - so we are not additional platform-depended on `nproc`...

RuffaloLavoisier:
I added the platformio compile job because the way I was doing plan it doesn't seem to work with Github. (Lets find out more...) Now it seems that the time to build all models will be shorter than before.

RuffaloLavoisier:
> * my PC it is already blazingly fast

:+1: 
@simonmicro 
But this is the Github CI system.

RuffaloLavoisier:
>What is the default of `-j`?

Control a number of parallel build jobs. Default is a number of CPUs in a system.
See https://docs.platformio.org/en/latest//core/userguide/cmd_run.html#cmdoption-pio-run-j

wait...does this mean the number of CPUs? Does that mean parallel work?

.github/buildEverything.py
```diff
@@ -72,16 +73,19 @@ def doBuild(makeDebug):
             configOut.close()
 
             # Compile firmware
-            logging.info('Compiling ' + filename + '...')
+            compile_start_time = datetime.now().strftime("%m/%d-%H:%M:%S")
+            logging.info('Compiling ' + compile_start_time + ' - ' + filename + '...')
+            job_number = str(os.cpu_count() * 2)
```
simonmicro, line 78: The hyper-threads are already counted into that number!

simonmicro (changes requested):

simonmicro:
> Default is a number of CPUs in a system

And I don't see any advantage of explicitly setting the CPU core count using `-j`...

RuffaloLavoisier:
Eureka... I think I found a way. I'll open up PR again.

RuffaloLavoisier:
See #221

RuffaloLavoisier:
In my humble opnion, I recommend working with one history or article. That way, we can check the previous changes and issues so you don're not confused. This PR has been disabled because a new PR was accidentally opened. I apologize.

PR #221, opened 2022-04-23
CI : update parallel compile from language resolve #217

### LINKED ISSUE #217: Improve the CI (FAST)

RuffaloLavoisier:


RuffaloLavoisier:
Oops.....there seems to be a problem with the submodule.

RuffaloLavoisier:
Amazing...I shortened it to 16 minutes...

RuffaloLavoisier:
In this way, we expect that all models can also be compiled in parallel to reduce to *1 minute* increments.  The downside is that you can't just automation-read the model or language and you have to input it every time it's updated or added.

RuffaloLavoisier:
Ddddd....why isn't the CI pipeline working?

simonmicro (changes requested):
My only request would be to add a `README.md` beside the locale headers, so any dev who adds a new language gets reminded to also extend the GitHub Actions pipeline.

simonmicro (approved):

PR #225, opened 2022-05-10
CI : find all languages and create a dynamic matrix

RuffaloLavoisier:
Dynamically explore all languages in the OSW main repository to create an array. This has the disadvantage of being able to see the results after the user adds and merges languages.

### DIFF
--- a/.github/workflows/test-OSW.yml
+++ b/.github/workflows/test-OSW.yml
@@ -7,12 +7,20 @@ on:
     branches: [ master, develop ]
 
 jobs:
+  find-languages:
+    runs-on: ubuntu-latest
+    steps:
+      - id: set-matrix
+        run: echo "::set-output name=languages_matrix::$(curl https://api.github.com/repos/open-smartwatch/open-smartwatch-os/contents/include/locales | jq -c '[.[].name[0:-2]]')"
+    outputs:
+      languages_matrix: ${{ steps.set-matrix.outputs.languages_matrix }}
   build_linux:
+    needs: find-languages
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        language: [ko-KR,cs-CZ,de-DE,en-US,fr-FR,hu-HU,it-IT]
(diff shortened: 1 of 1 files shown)


RuffaloLavoisier:
Importing the default branch...

.github/workflows/test-OSW.yml
```diff
@@ -7,12 +7,20 @@ on:
     branches: [ master, develop ]
 
 jobs:
+  find-languages:
+    runs-on: ubuntu-latest
+    steps:
+      - id: set-matrix
+        run: echo "::set-output name=languages_matrix::$(curl https://api.github.com/repos/open-smartwatch/open-smartwatch-os/contents/include/locales | jq -c '[.[].name[0:-2]]')"
```
simonmicro, line 14: You are referencing this repository using a hard-coded URI. Please change it to use some of the predefined GitHub Actions variables (https://docs.github.com/en/actions/learn-github-actions/environment-variables) and also to always refer to the files of the commit it is coming from.

I strongly advise you to use your Ubuntu shell of that job to parse the files of the repository from this GitHub Action run instead. It is local, does not need any external requests and should be more future-proof anyways.

simonmicro (changes requested):
See my comment.

.github/workflows/test-OSW.yml
simonmicro: Maybe use this bash-command instead?
```bash
cd include/locales/; X=$(ls *.h -1 | sed -e 's/\.h$//' | tr '\n' ' '); X=($X); jq --compact-output --null-input '$ARGS.positional' --args -- "${X[@]}"
```

This would be in YAML:
```yaml
- id: set-matrix
        run: echo "::set-output name=languages_matrix::$(cd include/locales/; X=$(ls *.h -1 | sed -e 's/\.h$//' | tr '\n' ' '); X=($X); jq --compact-output --null-input '$ARGS.positional' --args -- "${X[@]}")"
```

simonmicro (approved):

PR #229, opened 2022-05-31
New Release

simonmicro:


RuffaloLavoisier:
@simonmicro Why don't we change it so that the user can push([2256587](https://github.com/Open-Smartwatch/open-smartwatch-os/pull/229/commits/2256587a883cbde0c51e3d32479809aa67dc12a0)) themselves? What `CI` only does is to tell you the wrong part or leave a code review of the part that needs to be changed.

PR #236, opened 2022-06-11
Fallback Wi-Fi resolve #234

### LINKED ISSUE #234: Add Fallback Wi-Fi
I would like to add a Fallback-SSID (`cafe`, `school`...) so that Wi-Fi is connected even if it's not my main place(`home`). (*Security is yours!*)

*  It maybe doesn't need, but using AutoAP every time seems cumbersome.

RuffaloLavoisier:
* Add 2 more fallback Wi-Fi
* Optional Wi-Fi selection from the Web-interface


include/services/OswServiceTaskWiFi.h
```diff
@@ -66,7 +66,7 @@ class OswServiceTaskWiFi : public OswServiceTask {
     String m_clientSSID;
     String m_clientPass;
     String m_stationPass;
-
+    uint8_t count = 1;
```
uvwxy, line 69: `count` of what?

src/services/OswServiceTaskWiFi.cpp
```diff
@@ -179,15 +185,36 @@ bool OswServiceTaskWiFi::isWiFiEnabled() {
 void OswServiceTaskWiFi::connectWiFi() {
     this->m_hostname = std::move(OswConfigAllKeys::hostname.get());
     this->m_autoAPTimeout = 0; //First reset to avoid racing conditions
-    this->m_clientSSID = std::move(OswConfigAllKeys::wifiSsid.get());
-    this->m_clientPass = std::move(OswConfigAllKeys::wifiPass.get());
+
```
uvwxy, line 188: This is a lot of code added to the function -> extract to `loadCredentials(reconnectCount)`

uvwxy (changes requested):

RuffaloLavoisier (commented):
LGTM

src/services/OswServiceTaskWiFi.cpp
```diff
@@ -173,21 +179,44 @@ bool OswServiceTaskWiFi::isWiFiEnabled() {
     return this->m_enableClient;
 }
 
+void OswServiceTaskWiFi::loadCredentials(uint8_t reconnectCount) {
+  switch (reconnectCount) {
+    case 1:
+      this->m_clientSSID = std::move(OswConfigAllKeys::wifiSsid.get());
```
simonmicro, line 185: Do **NOT** use `std::move` here. It will **destroy** the data of the source (maybe you're lucky and Arduinos `String` class is too stupid for that, but...). DON'T.

simonmicro (changes requested):

include/config_defaults.h
```diff
@@ -77,7 +77,7 @@
 #ifndef CONFIG_FALLBACK_1ST_WIFI_PASS
 #define CONFIG_FALLBACK_1ST_WIFI_PASS ""
 #endif
-#ifndef CONFIG_FALLBACK_1ST_WIFI_PASS
+#ifndef CONFIG_FALLBACK_2ST_WIFI_PASS
```
RuffaloLavoisier, line 80: CONFIG_FALLBACK_2ST_WIFI_PASS -> CONFIG_FALLBACK_2ND_WIFI_PASS

simonmicro: -> 0a2c86155963c559eae14995ba939ee63a431c0b

PR #237, opened 2022-06-14
Apps : refactoring variable value

RuffaloLavoisier:
Modify some hardcoded values.

PR #239, opened 2022-06-14
OswAppWatchfaceMix : replace variable

RuffaloLavoisier:
See https://github.com/Open-Smartwatch/open-smartwatch-os/pull/237#discussion_r896417411


uvwxy:
👍

PR #242, opened 2022-06-15
HAL : add get last-week steps in environment

RuffaloLavoisier:
We can bring the data 7 days ago.

### DIFF
--- a/src/hal/environment.cpp
+++ b/src/hal/environment.cpp
@@ -176,13 +176,17 @@ uint32_t OswHal::Environment::getStepsToday() {
 }
 
 #ifdef OSW_FEATURE_STATS_STEPS
-uint32_t OswHal::Environment::getStepsOnDay(uint8_t dayOfWeek) {
+uint32_t OswHal::Environment::getStepsOnDay(uint8_t dayOfWeek, bool lastWeek) {
     uint32_t day = 0;
     uint32_t weekday = 0;
     OswHal::getInstance()->getLocalDate(&day, &weekday);
-    if (dayOfWeek == weekday)
+
+    if (!lastWeek and dayOfWeek == weekday) 
         return this->getStepsToday();
-    return this->_stepsCache[dayOfWeek];
+    else if(!lastWeek or (lastWeek and dayOfWeek == weekday))
+        return this->_stepsCache[dayOfWeek];
+    else
+        return 0; // In that case we don't have any history left anymore - just reply with a zero...
 }
 #endif
 uint32_t OswHal::Environment::getStepsTotal() {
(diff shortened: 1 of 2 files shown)


src/hal/environment.cpp
```diff
@@ -184,6 +184,12 @@ uint32_t OswHal::Environment::getStepsOnDay(uint8_t dayOfWeek) {
         return this->getStepsToday();
     return this->_stepsCache[dayOfWeek];
 }
+uint32_t OswHal::Environment::getStepsLastWeek() {
+    uint32_t currDoM = 0; // Unused, but required by function signature
+    uint32_t currDoW = 0;
+    OswHal::getInstance()->getLocalDate(&currDoM, &currDoW);
+    return this->_stepsCache[currDoW];
```
simonmicro, line 191: But... The function signature implies the step amount of the whole week, but instead you return the current days values (i know, theoretically this is the value from last week). I think you should change the function name or expand the `getStepsOnDay()` with a special value like `8` to access this instead.

simonmicro (changes requested):

src/hal/environment.cpp
RuffaloLavoisier: > I think you should change the function name

Under what name?

RuffaloLavoisier: Now name `getStepsLastWeek()`, If I were to change function name, what would it be? Am I right to understand?

simonmicro: Maybe `getStepsBefore7Days()`? I think I would prefer enhancing the aforementioned existing function.

Any input @uvwxy ?

RuffaloLavoisier: @simonmicro 
> I think I would prefer enhancing the aforementioned existing function.

So Is it right to do it with a special numbers `8`? Or Is it better to embed default parameters`getStepsOnDay(uint8_t day, bool Before7Days=false)`?

RuffaloLavoisier: If we are approaching with a special number, we need to change that much logic.
If we're trying to access it, the `for-loop`(Mainly accessible methods) that you usually use will have more unnecessary loops.
In my humble opinion, wouldn't the default parameters be the most appropriate?

simonmicro: The default parameter is also a great idea, but you won't be able to return any other step-value except for exactly the current day of last week? So, basically you have to return 0 or throw an exception for any other case... Still, this is a pretty good idea (best so far to solve this) - especially as it preserves the current interface design and allows for later expansion!

simonmicro (approved):

PR #244, opened 2022-06-15
Libraries : add sunset/sunrise time

RuffaloLavoisier:
This is a draft for sunrise/sunset composition.
See : #207

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -23,3 +23,6 @@
 	path = lib/LUA
 	url = https://github.com/lua/lua.git
 	branch = v5.3
+[submodule "lib/sunset"]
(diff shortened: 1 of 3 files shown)


simonmicro (commented):
You just added the lib. Fine. I mean, could you do anything with it? E.g. add it to a watchface, extend the HAL interface - anything? Otherwise the PR is fine, but sort of useless :sweat_smile:

RuffaloLavoisier:
@simonmicro 
Just make sure the things I added are correct I'm going to add an example after that.

RuffaloLavoisier:
Well, I personally think that this library is not useful. close it.

PR #246, opened 2022-06-22
CI check the unused flag

RuffaloLavoisier:
Added for check CI for unused functions.

### DIFF
--- a/.github/buildEverything.py
+++ b/.github/buildEverything.py
@@ -96,7 +96,17 @@ if __name__ == "__main__":
 
     if str(args["support_feature"]) != "":
         logging.info('Setting flag ' + str(args["support_feature"]) + '...')
-        os.system('sed -i'' -r -e "/build_flags/a\    -D '+str(args["support_feature"])+'" platformio.ini')
+        # Setup build flag (using the config file via replacing, as platformio does not allow setting the property using Python)
+        configIn = open(pioConfig, 'r')
+        configStr = configIn.read()
+        configIn.close()
+        configStr, hitCount = re.subn('build_flags =','build_flags =\n    -D '+str(args["support_feature"]),configStr)
+        if hitCount == 0:
+            logging.error('Error on setting build flag!')
+            exit(5)
+        configOut = open(pioConfig, 'w')
+        configOut.write(configStr)
+        configOut.close()
 
     #if you want all-language packs
     if args["support_language"] == "all" and args["support_model"] == "all":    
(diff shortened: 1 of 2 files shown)


feature-flag.md
```diff
@@ -0,0 +1,2 @@
+flag:RAW_SCREEN_SERVER
+flag:MATRIX
```
RuffaloLavoisier, line 2: We have to let them know that these flags exist somewhere.

.github/buildEverything.py
```diff
@@ -91,8 +91,13 @@ def doBuild(makeDebug):
 	
     ap.add_argument("-l", "--support-language", type=str, required=True, help="# model language to compile. (Enter 'all' to compile all language packs.)")
     ap.add_argument("-m", "--support-model", type=str, required=True, help="# model type to compile. (Enter 'all' to compile all model packs.)")
+    ap.add_argument("-f", "--support-feature", type=str, required=False, default="", help="# feature to compile. (Enter a feature to compile.)")
     args = vars(ap.parse_args())
 
+    if str(args["support_feature"]) != "":
+        logging.info('Setting flag ' + str(args["support_feature"]) + '...')
+        os.system('sed -i'' -r -e "/build_flags/a\    -D '+str(args["support_feature"])+'" platformio.ini')
```
RuffaloLavoisier, line 99: convert `Win cmd`?

RuffaloLavoisier (commented):
I think it needs to be corrected.

.github/buildEverything.py
simonmicro: Why did you ever use `os.system`? I think even here we used code to open the `platformio.ini` and to read it!

feature-flag.md
RuffaloLavoisier: Plan B, there's also a way to add a flag [here](https://github.com/Open-Smartwatch/open-smartwatch.github.io/blob/3357dbb630603e2e33255d887dbf3b32512fa566/docs/resources/firmware.md?plain=1#L5). I think we can do it.In my opinion, this is the most ideal.

.github/workflows/test-FEATURE.yml
```diff
@@ -0,0 +1,75 @@
...
+      uses: actions/setup-python@v2
+    - name: Install PlatformIO
+      run: python -m pip install --upgrade pip && pip install --upgrade platformio
+    
+    - name: Rename config
+      run: mv include/config.h.example include/config.h
+   
+    - name: Compile language ${{ matrix.language }} model ${{ matrix.model }} feature ${{ matrix.feature }}
+      run: python3 .github/buildEverything.py -l ${{ matrix.language }} -m ${{ matrix.model }} -f ${{ matrix.feature }}
+    - name: Upload firmware artifacts
```
simonmicro, line 70: I do not think we should archive the produced firmwares.

feature-flag.md
simonmicro: Maybe it is used - but I think the choice of `Markdown` for this config is pretty bad. Maybe take a look at YAML / JSON instead?

RuffaloLavoisier:
Current PR has resolved the current issue. I will open a new PR to update.

PR #248, opened 2022-06-24
Replace matrix to animation word

RuffaloLavoisier:
I changed it to an integrated meaning.


simonmicro:
I _personally_ think `ANIMATION` is too broad for **just** the Matrix animation, but whatever... Should be fine for now...

simonmicro (approved):

PR #250, opened 2022-06-29
CI : refactoring action with script

RuffaloLavoisier:
With #246

Replace some commands with a script.


RuffaloLavoisier (commented):
If there is anything else you need in the script part, please give me your opinion. otherwise LGTM.

PR #257, opened 2022-07-09
Allow changing button press timeouts via config.h

FranseFrikandel:
Maybe I am just exceptionally impatient, but the long button press timeouts just seemed too long to me, so I made the timeouts for both the long press of the stopwatch and app switcher configurable via config.h.


FranseFrikandel:
Ouch, that didn't work as expected! I'll open a new PR.

PR #258, opened 2022-07-09
Allow setting custom long press button timeouts

FranseFrikandel:
Maybe I am just exceptionally impatient, but the long button presses were taking too long for my liking, so I made them configurable via config.h.

### DIFF
--- a/include/config_defaults.h
+++ b/include/config_defaults.h
@@ -154,14 +154,19 @@
 #define THEME_DANGER_COLOR rgb888(255, 56, 96)
 #endif
 
+// Set the long press time to switch between apps.
+#ifndef DEFAULTLAUNCHER_LONG_PRESS
+#define DEFAULTLAUNCHER_LONG_PRESS 1000
+#endif
+
 // The following settings are configureable later on using the web ui, you can still set the defaults here.
 
-// Daylight saving time offset in hours. E.g. 0.5 = 30 min
 #ifndef CONFIG_DATE_FORMAT
 // possibilities: "yyyy.mm.dd" or "mm/dd/yyyy"
 #define CONFIG_DATE_FORMAT "mm/dd/yyyy"
 #endif
 
+// Daylight saving time offset in hours. E.g. 0.5 = 30 min
 #ifndef CONFIG_DAYLIGHTOFFSET
 #define CONFIG_DAYLIGHTOFFSET 0
 #endif
@@ -195,6 +200,9 @@
 #endif
 #ifndef TOOL_STOPWATCH
 #define TOOL_STOPWATCH 1
+#endif
+#ifndef TOOL_STOPWATCH_BTN_TIMEOUT
+#define TOOL_STOPWATCH_BTN_TIMEOUT 1800
 #endif
 #ifndef TOOL_WATERLEVEL
 #define TOOL_WATERLEVEL 1
--- a/include/osw_hal.h
+++ b/include/osw_hal.h
@@ -20,7 +20,6 @@
 #define ERR_SD_MISSING 1
 #define ERR_SD_MOUNT_FAILED 2
 
-#define DEFAULTLAUNCHER_LONG_PRESS 1000
 #define NUM_BUTTONS 3
 // enum for user space button handling
 enum Button { BUTTON_1 = 0, BUTTON_2 = 1, BUTTON_3 = 2 };
--- a/src/apps/main/stopwatch.cpp
+++ b/src/apps/main/stopwatch.cpp
@@ -42,7 +42,7 @@ void OswAppStopWatch::loop() {
     }
     long totalTime = diff + sumPaused;
 
-    long btnTimeout = 1800;
+    long btnTimeout = TOOL_STOPWATCH_BTN_TIMEOUT;
     long btnDown = 0;
     if(running) {
         if(hal->btnHasGoneDown(BUTTON_2)) {


src/apps/main/stopwatch.cpp
```diff
@@ -42,7 +42,7 @@ void OswAppStopWatch::loop() {
     }
     long totalTime = diff + sumPaused;
 
-    long btnTimeout = 1800;
+    long btnTimeout = TOOL_STOPWATCH_BTN_TIMEOUT;
```
RuffaloLavoisier, line 45: This is stopwatch not switch screens between apps..

RuffaloLavoisier (commented):
Hello world, thank you for your contribution. 

Does your opinion mean a long button time when switching screens between APP? 

But your has modified the stopwatch part. 

IMHO, I think this good idea.

I'd appreciate it if you could fix it. 

And isn't  this TimeOut(`1000`) introduced to prevent hardware [*debounce*](https://www.arduino.cc/en/Tutorial/BuiltInExamples/Debounce)? - @uvwxy 

Hey @FranseFrikandel, If you want to reflect this part, how about adding it to the `web-interface`? - @simonmicro

FranseFrikandel:
I had changed both the long press to reset the stopwatch, and the long press to switch between apps. If you prefer I could change the PR to be only the one between switching apps.

Definetly willing to look into implementing it into the web interface, but I haven't gotten WiFi on my watch working yet, will get back to that this evening.

RuffaloLavoisier:
@FranseFrikandel 
I'm sorry, I didn't quite understand.

What do you want do?
Does it mean that the screen switching time between apps is long when you press the button? Or does it mean that the initialization time of the stopwatch is long?

For your information, your PR is modified the initialization time of the stopwatch.

FranseFrikandel:
The PR makes the long press time configurable for both app switching and resetting the stopwatch, as both of these were too long for my liking.

My changes in include/osw_hal.h are for customizable long press times to switch between apps (bottom left button). This was currently hardcoded to 1000 milliseconds, and reflected in the variable DEFAULTLAUNCHER_LONG_PRESS. The default remains the same as before at 1000 milliseconds but is now editable via changing DEFAULTLAUNCHER_LONG_PRESS in your config.h

The changes to stopwatch.cpp are the long press to reset the stopwatch once you have paused it (bottom right button). (so setting it back to 0) This was currently hardcoded to 1800ms and is now also configurable via the new variable TOOL_STOPWATCH_BTN_TIMEOUT. This variable also still defaults to its original 1800ms just as the switching between apps before.

So as an example, in my config.h I have now set this:
`
#define TOOL_STOPWATCH_BTN_TIMEOUT 800
#define DEFAULTLAUNCHER_LONG_PRESS 600
`

Which makes resetting the stopwatch take 800ms, and switching between apps 600ms.

So to answer both questions directly:
1. The time I had to hold the button down to switch between apps instead of switching between pages in apps was too long for my liking and adressed in this PR.
2. Not the initialization of the stopwatch, but the time you had to hold down the reset button of the stopwatch to reset the counter back to 0.

Hope this clarifies everything.

> @FranseFrikandel I'm sorry, I didn't quite understand. There is not enough explanation for PR. Would you explain in detail? 😆
> ### What do you mean?
> 
>     1. Does it mean that the screen switching time between apps is long when you press the button?
> 
>     2. Does it mean that the initialization time of the stopwatch is long?
> 
> 
> For your information, your PR is 2(initialization time of the stopwatch).

include/osw_hal.h
```diff
@@ -20,7 +20,6 @@
 #define ERR_SD_MISSING 1
 #define ERR_SD_MOUNT_FAILED 2
 
-#define DEFAULTLAUNCHER_LONG_PRESS 1000
```
RuffaloLavoisier, line 23: Great ! 👍

src/apps/main/stopwatch.cpp
RuffaloLavoisier: This can be changed to Web-interface.

include/config_defaults.h
```diff
@@ -196,6 +201,9 @@
 #ifndef TOOL_STOPWATCH
 #define TOOL_STOPWATCH 1
 #endif
+#ifndef TOOL_STOPWATCH_BTN_TIMEOUT
```
RuffaloLavoisier, line 204: Can u please revert this ?

include/config_defaults.h
```diff
@@ -154,14 +154,19 @@
...
+
 // The following settings are configureable later on using the web ui, you can still set the defaults here.
 
-// Daylight saving time offset in hours. E.g. 0.5 = 30 min
 #ifndef CONFIG_DATE_FORMAT
 // possibilities: "yyyy.mm.dd" or "mm/dd/yyyy"
 #define CONFIG_DATE_FORMAT "mm/dd/yyyy"
 #endif
 
+// Daylight saving time offset in hours. E.g. 0.5 = 30 min
```
RuffaloLavoisier, line 169: Thanks !

include/config_defaults.h
```diff
@@ -154,14 +154,19 @@
 #define THEME_DANGER_COLOR rgb888(255, 56, 96)
 #endif
 
+// Set the long press time to switch between apps.
+#ifndef DEFAULTLAUNCHER_LONG_PRESS
+#define DEFAULTLAUNCHER_LONG_PRESS 1000
+#endif
+
```
RuffaloLavoisier, line 161: Can u please revert this ?

RuffaloLavoisier (changes requested):
Thank you for the explanation. All right, I understand.
I think it would be better to use the web interface.

Could you please refer to the comments left and restore the parts that can be returned to the past commit? (remove)

And see this commit (96f3ae0c5e44a780b1937625933b55ba3956613a) if you want to know how the Web interface allows users to access the Web server and change the value.

 If you need help, hit (weak..) mention 'Ruffalo' in Discord.

FranseFrikandel:
I will get to these changes once I get the WiFi working on my watch 😅.

simonmicro:
I think the idea is great and dead-simple to be implemented. @RuffaloLavoisier you should keep in mind, that any variable accessible inside the web interface can also be default-initialized using the `config.h` header file. I'll just slap something together quickly and push it into the forks branch. Just a minute :)

simonmicro:
Ah shit. I f*ed this PR - GitHub seems to close it if I remove all commits?! **I'm very sorry - I'll reopen an other PR as a replacement from my fork!**

@FranseFrikandel Sorry, could you checkout #262

PR #263, opened 2022-07-11
Bug/esp fleeing time

### LINKED ISSUE #249: Time running fast
# What is not working?
I noticed the time on the watch running fast.
When synchronizing the time using wifi, the time is set correctly again.
On the debug screen I noticed 2 values for the seconds. The ESP32 value and the DS3231MA.
The ESP32 value is the one running fast.

# How to reproduce this
- Synchronize the time using wifi.
- Wait for any amount of time. The longer you wait, the longer the time difference will be.

# What should happen instead?
- I expect the time to not run fast.
- I expected the time to be read from the DS3231MA / be synchronized with the DS3231MA value more.

# Environment
- **Model** Light Edition [e.g. Light Edition]
- **OSW-OS** Code cloned from master branch 19-06-2022

# Additional context
This issue started after having uploaded a new version of the OSW software.
Since this was not an issue with previous software versions, I do not think this is a hardware issue.

simonmicro:
Fixes #249 - hopefully?

### DIFF
--- a/include/devices/esp32.h
+++ b/include/devices/esp32.h
@@ -10,7 +10,7 @@ class NativeESP32 : public OswTemperatureProvider, public OswTimeProvider {
     virtual ~NativeESP32() {};
 
     virtual void setup() override;
-    virtual void update() override {};
+    virtual void update() override;
     virtual void reset() override {};
     virtual void stop() override {};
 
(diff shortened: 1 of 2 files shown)


RuffaloLavoisier:
Can't we fix it with periodic NTP updates?

simonmicro:
> Can't we fix it with periodic NTP updates?

Yes, for that you can just use the "auto updatr on wifi" inside the webgui-setting. But this wn't work if the watch just started and has no network (yet). So this just resyncs with the extra RTC...

PR #266, opened 2022-07-13
OSW-settingBrightness : refactoring and add in debug part

RuffaloLavoisier:


simonmicro (approved):

RuffaloLavoisier (commented):
LTGM

RuffaloLavoisier:
control_diplay_osw

Brightness icon.

src/apps/watchfaces/OswAppWatchfaceMix.cpp
```diff
@@ -93,23 +93,37 @@ void OswAppWatchfaceMix::digitalWatchDisplay() {
...
-    if (hal->btnHasGoneDown(BUTTON_3)) {
-        hal->increaseBrightness(25);
+    if (hal->btnIsDown(BUTTON_3) && hal->btnHasGoneDown(BUTTON_2)) {
+        buttonMode = (buttonMode + 1) % 2;
     }
-    if (hal->btnHasGoneDown(BUTTON_2)) {
-        hal->decreaseBrightness(25);
+    ui->watchfaceBrightnessControlMode = buttonMode;
+
+    switch (buttonMode) {
```
simonmicro, line 107: This would require every app to also have this `switch`-statement in it by itself. I think this will make future extension pretty hard! How about you create a static method inside the watchface / UI class, which handles the buttons and expects e.g. a callback for an additional button mode called "app-specific" or "other" or "custom". If the user selects this mode (own icon?), the callback is just being called whenever the static method is being executed from the app.

src/osw_ui.cpp
```diff
@@ -101,6 +101,9 @@ void OswUI::loop(OswAppSwitcher& mainAppSwitcher, uint16_t& mainAppIndex) {
             // Only draw on first face if enabled, or on all others
             if ((mainAppIndex == 0 && OswConfigAllKeys::settingDisplayOverlaysOnWatchScreen.get()) || mainAppIndex != 0)
                 drawOverlays();
+        if (mainAppIndex == 0 && watchfaceBrightnessControlMode == 0) {
```
simonmicro, line 104: :exclamation: This belongs into the overlay function `drawOverlays()`. Not here.

PR #271, opened 2022-07-20
Fixed Lightsleep not working properly 

Gladox114:
Fixed Lightsleep not working properly with really really long enough button press as it switches to the next app because doSwitch is still true form LONG_PRESS.

### DIFF
--- a/src/apps/main/switcher.cpp
+++ b/src/apps/main/switcher.cpp
@@ -39,13 +39,16 @@ void OswAppSwitcher::loop() {
         }
     }
 
+    bool sleeping = false;
+
     // do action only once the button goes up
     if (hal->btnHasGoneUp(_btn)) {
         if (_doSleep) {
             _doSleep = false;
+            _doSwitch = false; // if lightsleep is enabled then doSwitch is still true and will be used the next time.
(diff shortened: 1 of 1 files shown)


src/apps/main/switcher.cpp
```diff
@@ -43,9 +43,10 @@ void OswAppSwitcher::loop() {
     if (hal->btnHasGoneUp(_btn)) {
         if (_doSleep) {
             _doSleep = false;
+            _doSwitch = false; // if lightsleep is enabled then doSwitch is still true and will be used the next time.
+            sleeping = true; // because in lightsleep the next _app will use the button and switch the app which isn't what we want
```
RuffaloLavoisier, line 47: Not declared, and it should prevent initialization when it wake up.

RuffaloLavoisier (changes requested):

src/apps/main/switcher.cpp
Gladox114: Whoops I forgot to add the sleeping variable initialization as I worked with two different forks and just drag and dropped my changes.

Gladox114:
I just reopen it as I don't know how to do it properly.

RuffaloLavoisier:
@Gladox114 
If you have any difficulties, ping, don't waste your pointless PR .

PR #297, opened 2022-09-27
Fix warnings

jessp01:
Fix `ISO C++ says that these are ambiguous, even though the worst conversion for the first is better than the worst conversion for the second` warnings.


RuffaloLavoisier (changes requested):
Thank you for your contribution and finding the error :+1: 
Could you briefly write in the message what you fixed in the code? What have you improved or modified?
We want to write the key points while simplifying the commit message for changes or for future maintenance.

uvwxy:
Thank you @jessp01 !

If you squash the commits into one (as the last two are merely cleanup of the first), this should be ready to be merged.

(I'm not familiar enough with c++ at this point to understand the warning, but maybe this led to this in the first place ;) )

If you need any input to do this, please let us know. 👍

simonmicro:
@jessp01 Hey, thanks for patching this. I've not done that myself, as I wanted to squash the problem at its root and note by merely work-arounding the symptoms. I'll take a look into this by myself again, but maybe we'll just go with this instead.

src/osw_config.cpp
```diff
@@ -143,7 +143,9 @@ void OswConfig::parseDataJSON(const char* json) {
                 break;
             }
         if (!key) {
-            Serial.println("WARNING: Unknown key id \"" + entryId + "\" provided -> ignoring...");
+            Serial.print("WARNING: Unknown key id \""); 
+	    Serial.print(entryId);
```
simonmicro, line 147: Wrong indentation.

simonmicro (changes requested):

simonmicro:
And yes, as @uvwxy already mentioned - whoever merges this: Please squash the commits!

_Maybe this is a note to my future-self, which will 100% ignore this..._

simonmicro:
And therefore 97e0f200c3843b39467ad172afee609b47200cc1 is the proper fix. Thanks for your attempt to fix this, it was just at the wrong place. :grin:

PR #288, opened 2022-09-03
Add Monotimer Watchface

daschr:
Hi!

I've created a monotimer watchface.

Would be nice if you review and (maybe) merge it.

Cheers,

David

PR #298, opened 2022-09-27
Numeral watchface

jessp01:
Hello all,

Well done on this excellent project.

I've created this watch face:

numerals

Would be glad to see it merged.

Cheers,


RuffaloLavoisier (changes requested):
Since we have a similar series watch, why don't we add the choose whether to display numbers in the web interface?

uvwxy:
> Since we have a similar series watch, why don't we add the choose whether to display numbers in the web interface?

I would be in favour of merging this as a separate watchface. As this uses additional fonts these would be added to the binary size of the OS even if the setting is not used.

By not including the watchfaces at compile time we can exclude additional, watchface specific dependencies at compile time, too.

RuffaloLavoisier:
> > Since we have a similar series watch, why don't we add the choose whether to display numbers in the web interface?
> 
> I would be in favour of merging this as a separate watchface. As this uses additional fonts these would be added to the binary size of the OS even if the setting is not used.
> 
> By not including the watchfaces at compile time we can exclude additional, watchface specific dependencies at compile time, too.

As the Watchface increases, it takes a long time to find or manage the design you want, so you need to find a way to solve it.

RuffaloLavoisier (changes requested):
LGTM, Additionally, you must add this clock to the Web interface index.

RuffaloLavoisier:
Corrected commit log.

jessp01:
Hi all,

I didn't want to make changes to other watchface files without discussing it with you (the maintainers) but generally speaking, I'd be all for the below modifications and am happy to submit them if everyone agrees:
- One function to draw the numerals; this can accept the font to use (of course, we have to limit the selection to suitable fonts, can't use a 24pt one for example), coordinates, etc as parameters
- Expose a configuration that draws the numerals conditionally, which can then be checked in applicable watchfaces

If that sounds reasonable, which file should we include the numeral drawing function in? perhaps under `include/apps/watchfaces/utils`?

Additionally, I thought of adding the day of the month like so:
image

Lastly, @RuffaloLavoisier,
> Sometimes this step-counter will not be readable by the user by the second hand of the clock.

Indeed, but I reckon that's true to many different designs, including the one in `OswAppWatchfaceMonotimer` (although far less frequently) and other smart watch projects as well. I can use a smaller font size for it but personally, I don't think it's too bad as is. 
The same statement applies to displaying the day of the month, of course.

> LGTM, Additionally, you must add this clock to the Web interface index.
Just so I'm sure, do you mean I should add it to https://github.com/Open-Smartwatch/open-smartwatch.github.io/blob/source/docs/index.md?

I'd also be open to contributing documentation on how to add watchfaces, which could help newcomers.

Thanks,

RuffaloLavoisier:
LGTM, I'm happy to merge.

> I'd also be open to contributing documentation on how to add watchfaces, which could help newcomers.
> 
> Thanks,

https://github.com/Open-Smartwatch/open-smartwatch-os/pull/288/files#diff-1eb7e6ee52c2add1ca7bafdf476d09de534e15be005de55cd51f235cc9f48c75R47

Please refer to this part And please write a brief introduction to this watch because we need to add it to the document (how does the number of steps look, the number of times of the week, etc.).

include/apps/watchfaces/OswAppWatchfaceNumerals.h
```diff
@@ -0,0 +1,32 @@
...
+    ~OswAppWatchfaceNumerals() {};
+
+#ifdef OSW_FEATURE_STATS_STEPS
+    static void drawStepHistory(OswUI* ui, uint8_t x, uint8_t y, uint8_t w, uint8_t h, uint32_t max);
+#endif
+  private:
+    OswUI* ui;
+    void drawWatch();
+    void drawNShiftedTicks(Graphics2D* gfx, uint8_t cx, uint8_t cy, uint8_t r1, uint8_t r2, uint8_t nTicks, float shift, uint16_t color);
+    void drawNShiftedMaskedTicks(Graphics2D* gfx, uint8_t cx, uint8_t cy, uint8_t r1, uint8_t r2, uint8_t nTicks, float shift, uint16_t m, uint16_t color);
```
RuffaloLavoisier, line 28: not used :)

src/apps/watchfaces/OswAppWatchfaceNumerals.cpp
```diff
@@ -0,0 +1,138 @@
...
+        21, 120,    // 09
+        37, 80,     // 10
+        65, 45,     // 11
+        110, 30     // 12
+    };
+
+    for (uint8_t i=0; i<12; ++i) {
+        hal->gfx()->setTextCursor(positions[i<<1], positions[(i<<1)|1]);
+        hal->gfx()->printDecimal(i+1, 2);
+    }
```
RuffaloLavoisier, line 54: This will be done by reusing the monotimer part.

RuffaloLavoisier (changes requested):
Most of the time, the Watchface uses a variety of similar features and another watchfaces. Therefore, it is better to choose how to reuse the code rather than duplicate it.

And please update display date part :)

include/apps/watchfaces/OswAppWatchfaceNumerals.h
jessp01: Indeed, removed:)

RuffaloLavoisier: Yeah i see 👍 
I mean, please delete this code.

jessp01:
Hi @RuffaloLavoisier,

Sorry, I've neglected to commit the changes I made to `src/osw_config_keys.cpp` before. Did so now.
Shall I also submit a pull to update https://github.com/Open-Smartwatch/open-smartwatch.github.io/blob/source/docs/index.md?

> Most of the time, the Watchface uses a variety of similar features and another watchfaces. Therefore, it is better to choose how to reuse the code rather than duplicate it.
I agree but to me, it feels less desirable to rely on code implemented in other watchfaces which is why I made the suggestions in my last comment (https://github.com/Open-Smartwatch/open-smartwatch-os/pull/298#issuecomment-1262000896). Please let me know what you think.

> And please update display date part :)
Updated. This too can be placed in a util function that can be then used in the different watch faces as per my last comment. Happy to refactor if you find this acceptable.

Thanks,

RuffaloLavoisier:
Your commit is based on the `master`. 
This merge builds up unnecessary commit logs. 
I got it corrected commit log, but you force-push...  :laughing: 
Please use, if you need..
```
$ git fetch && git pull
```

RuffaloLavoisier:
> which is why I made the suggestions in my last comment ([#298 (comment)](https://github.com/Open-Smartwatch/open-smartwatch-os/pull/298#issuecomment-1262000896)).

That comment is about the overlap between the number of steps and the hands of the clock. This is a (https://github.com/Open-Smartwatch/open-smartwatch-os/pull/298#discussion_r983517852) problem that makes code reuse, or repetitive code, difficult to maintain later

RuffaloLavoisier:
Corrected the commit log again...
```
$ git pull
```

RuffaloLavoisier (approved):
LGTM - I've modified and organized some codes. There are more things to organize, but I think it would be okay to do it when I organize the watchface later.

RuffaloLavoisier:
@jessp01 
First, I'm sorry and thank you. It is sometimes good to modify internally like this. 
As a result, thanks to your idea, we are able to merge the new watchface. 
Thank you for your contribution.

RuffaloLavoisier (approved):

PR #308, opened 2022-10-07
Block drift compensation when NTP time update is in progress

### LINKED ISSUE #307: NTP time updating doesn't work due to time drift compensation.
# What is not working?
PR #263 appears to break NTP network update. The time is not updated when pressing the update button when connected to WiFi.

It appears that immediatly as the time is set, the automatic time drift detection pulls the time back from the onboard RTC, prevent time from being effectively updated. Commenting out the code that keeps the ESP32 internal clock in sync appears to fix the issue as well (line 42 to line 50 in src/devices/esp32.cpp).

# How to reproduce this
Try to update the time using NTP when it is significantly off.

# What should happen instead?
The time should update.

# Environment
- OSW Light v3.2
- OSW latest main branch

# Additional context
Serial monitor:
```
src/services/OswServiceTaskWiFi.cpp: [NTP] Started update...
src/devices/esp32.cpp: Resynced internal ESP32 clock with other provider due to significant time difference (> 4 seconds).
src/services/OswServiceTaskWiFi.cpp: [NTP] Update finished (time of 1665131972)!
```

FranseFrikandel:
Blocking the NTP time compensation while NTP time update is in progress seems to resolve #307.

I think the following is what was happening:
The ESP32's time gets set to 0 to signify the time is not yet updated inside the triggerNTPUpdate function of esp32.cpp.
When the drift compensation sets the time again to the same time as is saved in the RTC before the NTP update is finished, the code in OswServiceTaskWiFi.cpp at line 107 gets triggered prematurely. This means the time of the ESP gets set back to the same time as that of the RTC before the NTP update was ever able to complete.

For some reason github is adding a bunch of my old commits that got somewhat jumbled around in my last PR. The changed files that are listed are correct, but some of the commits it adds shouldn't be there, however they seemingly are overwritten anyway. I don't know how to fix this.


RuffaloLavoisier:
It's still working.
@simonmicro Can you please check it?

FranseFrikandel:
I see I still need to make sure the emulator compiles, didn't think to check that. Will probably be monday when I get back to my desktop PC.

Also it might not happen always as it is a race condition the way I see it, but it was happening quite consistently to me. I did lower the amount of memory that triggers the low memory state to 20kB rather than 100, so maybe not being in low memory state and the slowdown associated with it causes it to happen more often?

simonmicro:
Sorry, didn't see you #308 ! I was just sitting on a train-ride and saw the issue note in our Discord. I thought "quick fix" and implemented it myself. If OK with you, I would prefer to merge #309 instead this one.

PR #310, opened 2022-10-10
Add NTP status message

EliRoberts5:
When the network time is updated via Button 2 , a message is displayed while the update takes place. An NTP status method is added. Some minor BLE changes are added


include/locales/en-US.h
```diff
@@ -11,6 +11,7 @@
 #define LANG_MANUALLY "Manually"
 #define LANG_CANCEL "Cancel"
 #define LANG_SAVE "Save"
+#define LANG_NTP_QUEUED "NTP Updating"
```
RuffaloLavoisier, line 14: Setup is required for all languages that currently exist.

include/services/OswServiceTaskWiFi.h
```diff
@@ -44,6 +44,8 @@ class OswServiceTaskWiFi : public OswServiceTask {
     const String& getStationPassword() const;
 
     ~OswServiceTaskWiFi() {};
+     // ** updated added NTP status method to return NTP status 
+    bool getNTPstatus(void);
```
RuffaloLavoisier, line 48: Just comment `// blah blah`.
And `bool getNTPstatus();`

src/services/OswServiceTaskWiFi.cpp
```diff
@@ -209,6 +209,13 @@ bool OswServiceTaskWiFi::isConnected() {
     return this->m_enableWiFi and (this->m_enableStation or WiFi.status() == WL_CONNECTED);
 }
 
+// ** updated NTP status thing paul october 07 2022
+bool OswServiceTaskWiFi::getNTPstatus(void){
+    
+    return this->m_waitingForNTPUpdate;
```
RuffaloLavoisier, line 215: @simonmicro 
I think we need an update on this.

src/apps/tools/OswAppTimeConfig.cpp
```diff
@@ -180,9 +180,21 @@ void OswAppTimeConfig::loop() {
         if (OswServiceAllTasks::wifi.isConnected()) {
             OswUI::getInstance()->setTextCursor(BUTTON_2);
             hal->gfx()->print(LANG_TFW_UPDATE);
+            // ** if time is still being updated reprint this status on the display
+            if (OswServiceAllTasks::wifi.getNTPstatus()  ) {    
+                hal->gfx()->setTextCursor(210,150);
+                hal->gfx()->print(LANG_NTP_QUEUED);  // ** NTP Updating message
+            }
```
RuffaloLavoisier, line 187: Maybe
```
// if time is still being updated reprint this status on the display
if (OswServiceAllTasks::wifi.getNTPstatus()  ) {    
    hal->gfx()->print(LANG_NTP_QUEUED);  // ** NTP Updating message
} else {
    hal->gfx()->print(LANG_TFW_UPDATE);
}
```

src/apps/tools/OswAppTimeConfig.cpp
```diff
@@ -180,9 +180,21 @@ void OswAppTimeConfig::loop() {
...
             if (hal->btnHasGoneDown(BUTTON_2)) {
                 if (OswServiceAllTasks::wifi.isConnected()) {
                     OswServiceAllTasks::wifi.queueTimeUpdateViaNTP();
+                    // ** when time update button2 is pressed, display NTP updating message
+                    // ** until update is complete
+                    hal->gfx()->setTextCursor(210,150);
+                    hal->gfx()->print(LANG_NTP_QUEUED);  // ** NTP Updating message
+                    #ifndef NDEBUG
+                        Serial.println("updating time via NTP");
+                    #endif
```
RuffaloLavoisier, line 197: Please remove it.

RuffaloLavoisier (changes requested):
Thank you for your contribution.

We are currently developing the BLE part, so please exclude this part separately.
So please remove `BLE` and `PIO` config file from git history?

@simonmicro 
Is it worth merging only NTP update notifications into `bugs/tooAggressiveTimeSync` branch?

include/locales/en-US.h
EliRoberts5: ok, unless there is another way to do this, I would be utilizing Google translate for all languages other then english and french.

include/services/OswServiceTaskWiFi.h
EliRoberts5: can you be more specific, do you request that the comment be removed?

RuffaloLavoisier: What is that `**`?  ..  😄

src/apps/tools/OswAppTimeConfig.cpp
EliRoberts5: ok, I will look at it ..

include/services/OswServiceTaskWiFi.h
RuffaloLavoisier: And Just `bool getNTPstatus();` not `bool getNTPstatus(void);`.
For the other things and the rules.

EliRoberts5: ok you want the "void" removed ..

src/apps/tools/OswAppTimeConfig.cpp
EliRoberts5: you want the message removed?

RuffaloLavoisier: remove code.

EliRoberts5: ok, it should still work without that code, I will have to test it to see, then I will remove it if that is the case. good call I think, although the user will not get any indication when they actually press the button, only when the update is still in progress. If you prefer this, I would appreciate if you would confirm.

EliRoberts5: I also have a fairly large update for the watchfaces, wondering if you have any comment on how to submit those changes?

EliRoberts5:
I am going to redo this and resubmit as a new PR.

Issue #168, opened 2021-11-03
Unify console handling & logging


simonmicro:
All the logging should be centrally done through a thread-save system (including source, severity information) - also allowing us to buffer messages to show them inside e.g. a web console. Also add some commands to interact with the watch - e.g. to set the SSID/PWD or... Well, the time.

Maybe even include some form of...
```
The programs included with the OSW-OS system are free software;
the exact distribution terms for each program are described in the
individual files.

OSW-OS comes with ABSOLUTELY NO WARRANTY, to the extent
permitted by applicable law.
```

PR #312, opened 2022-10-12
Unified logging for #168

simonmicro:
* Introduces `OSW_LOG_D/I/W/E()` for #168...
* Will **drastically** shrink release builds (+speedup), as debug-messages are fully removed (even their concat-operations)
* Will somewhat reduce debug builds, as e.g. `__FILE__` is now clearer just a static pointer, which can be deduplicated by the compiler

- [x] Also fixed minor log message copy-pasta-bug inside the environment-init and sensor unavailable warnings...
- [x] Test on real hardware for crashes...
- [x] ...and correct datatype->message conversion
- [x] Handle multiple `\n` in logging messages


src/hal/environment.cpp
```diff
@@ -18,50 +18,40 @@ void OswHal::Environment::updateProviders() {
...
     for(auto& d : *OswAccelerationProvider::getAllAccelerationDevices()) {
         if(this->accelSensor == nullptr or this->accelSensor->getAccelerationProviderPriority() < d->getAccelerationProviderPriority())
             this->accelSensor = d;
     }
-#ifndef NDEBUG
-    if(!this->tempSensor)
-        Serial.println(String(__FILE__) + ": Acceleration API enabled, but no provider available!");
-#endif
+    if(!this->accelSensor)
+        OSW_LOG_W("Acceleration API enabled, but no provider available!");
```
RuffaloLavoisier, line 30: In this case, the conditional statements are always compiled and always passed through the conditional statements, with or without debug flags. Previously, if statements were compiled only when the NDEBUG flag was declared. Can this also be included as a function or condition to output the log?
Use default param??
```
OSW_LOG_W("Acceleration API enabled, but no provider available!",!this->accelSensor);
```
like this...

src/hal/environment.cpp
```diff
@@ -18,50 +18,40 @@ void OswHal::Environment::updateProviders() {
         if(this->tempSensor == nullptr or this->tempSensor->getTemperatureProviderPriority() < d->getTemperatureProviderPriority())
             this->tempSensor = d;
     }
-#ifndef NDEBUG
     if(!this->tempSensor)
-        Serial.println(String(__FILE__) + ": Temperature API enabled, but no provider available!");
-#endif
+        OSW_LOG_W("Temperature API enabled, but no provider available!");
```
RuffaloLavoisier, line 22: same..

include/debug_scani2c.h
```diff
@@ -20,24 +21,17 @@ void scanI2C() {
...
+            OSW_LOG_E("Unknown error at address 0x", address < 16 ? "0" : "", String(address, HEX));
         }
     }
     if (nDevices == 0)
-        Serial.println("No I2C devices found\n");
+        OSW_LOG_I("No I2C devices found");
     else
-        Serial.println("done\n");
-
-    delay(5000);  // wait 5 seconds for
```
RuffaloLavoisier, line 40: Where did this go?

include/debug_scani2c.h
```diff
@@ -20,24 +21,17 @@ void scanI2C() {
...
-            if (address < 16) Serial.print("0");
-            Serial.println(address, HEX);
+            OSW_LOG_E("Unknown error at address 0x", address < 16 ? "0" : "", String(address, HEX));
         }
     }
     if (nDevices == 0)
-        Serial.println("No I2C devices found\n");
+        OSW_LOG_I("No I2C devices found");
     else
-        Serial.println("done\n");
```
RuffaloLavoisier, line 38: Why did you add `\n` to the output when you already have the `println` function?

include/debug_scani2c.h
```diff
@@ -20,24 +21,17 @@ void scanI2C() {
...
-            Serial.print(address, HEX);
-            Serial.println("  !");
+            OSW_LOG_I("I2C device found at address 0x", address < 16 ? '0' : '', msg, String(address, HEX), " !");
 
             nDevices++;
         } else if (error == 4) {
-            Serial.print("Unknown error at address 0x");
-            if (address < 16) Serial.print("0");
-            Serial.println(address, HEX);
+            OSW_LOG_E("Unknown error at address 0x", address < 16 ? "0" : "", String(address, HEX));
```
RuffaloLavoisier, line 28: Maybe I think it's kind of like this.

>In this case, the conditional statements are always compiled and always passed through the conditional statements, >with or without debug flags. Previously, if statements were compiled only when the NDEBUG flag was declared. Can >this also be included as a function or condition to output the log?
>Use default param??
>```
>OSW_LOG_W("Acceleration API enabled, but no provider available!",!this->accelSensor);
>```
>like this...

include/OswLogger.h
```diff
@@ -0,0 +1,120 @@
...
+        if(severity == severity_t::D)
+            Serial.print("D: ");
+        else if(severity == severity_t::I)
+            Serial.print("I: ");
+        else if(severity == severity_t::W)
+            Serial.print("W: ");
+        else if(severity == severity_t::E)
+            Serial.print("E: ");
+        else
+            throw std::logic_error("Unknown severity level");
```
RuffaloLavoisier, line 97: How about `switch-case`

RuffaloLavoisier (changes requested):

src/hal/environment.cpp
simonmicro: No, the statements are not compiled in builds, which have the `NDEBUG`-flag. Because in those builds, the `OSW_LOG_W` will be expanded into nothing. So practically there is written: `if(!this->accelSensor);`, which will be removed by the compiler, as the condition does not cause any changes to the object / state.

include/debug_scani2c.h
simonmicro: This is a good question... @uvwxy :P

include/debug_scani2c.h
simonmicro: As far as I can tell this sleep has no purpose today anymore and just slows down rapid scanning. So I removed it.

include/debug_scani2c.h
simonmicro: The interface of the `OswLogger` was explicitly designed to remove the need for string-concatenation. This line is an example how to use it... I don't think putting conditions into potentially never expanded defines is a good idea. Take a look into `assert()` - this function is exactly one of these dangers, as sometimes you put relevant code into its condition, which won't be compiled into release-builds. Very uncool :grin:

include/OswLogger.h
simonmicro: In Kotlin we could talk about that, but in C++? No... This is one of these classic topics :roll_eyes: . I think switch-case is only useful if you want to use its fall-trough feature, in any other case the danger of missing only one `break` is too high - causing unpredictable behavior. Ofc we could discuss, that I even omitted the `{}`, but that will be fixed by the auto-formatter anyways...

**BUT!** I've just read this article: https://www.scaler.com/topics/c/difference-between-if-else-and-switch/
They mention the creation of jump-tables, which will improve performance :star_struck: 

Soooo... Proposal: Let this one pass for now and I'll create a PR down the road to migrate as-much-as-possible to `switch`-statements inside the whole OS.

PR #313, opened 2022-10-16
CI: skip unnecessary actions

RuffaloLavoisier:
PR for branches at the same storage level takes a lot of time to check with one push and one PR. I think we can reduce this in this way.

### DIFF
--- a/.github/workflows/test-EMULATOR.yml
+++ b/.github/workflows/test-EMULATOR.yml
@@ -14,7 +14,7 @@ jobs:
       should_skip: ${{ steps.skip_check.outputs.should_skip }}
     steps:
       - id: skip_check
-        uses: fkirc/skip-duplicate-actions@master
+        uses: fkirc/skip-duplicate-actions@v5.2.0
         with:
           concurrent_skipping: 'same_content_newer'
           skip_after_successful_duplicate: 'true'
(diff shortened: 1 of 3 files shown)


RuffaloLavoisier:
I've improved this. https://github.com/Open-Smartwatch/open-smartwatch-os/pull/301

RuffaloLavoisier:
LGTM

.github/workflows/test-EMULATOR.yml
```diff
@@ -3,14 +3,27 @@ name: OSW-EMULATOR-test
 on:
   workflow_dispatch:
   push:
+    paths_ignore: '["**/*.md", "**/scripts/screen_capture/**"]'
```
simonmicro, line 6: I think the screen capture path is somewhat very specific - but I have no idea to make it better :)

simonmicro (approved):

.github/workflows/test-EMULATOR.yml
```diff
@@ -3,14 +3,27 @@ name: OSW-EMULATOR-test
...
     branches: [ master, develop ]
 
 jobs:
+  check_skip:
+    runs-on: ubuntu-latest
+    outputs:
+      should_skip: ${{ steps.skip_check.outputs.should_skip }}
+    steps:
+      - id: skip_check
+        uses: fkirc/skip-duplicate-actions@master
```
simonmicro, line 17: Do not use master here - please pin it to a specific version number or even commit hash instead!

simonmicro (changes requested):

PR #315, opened 2022-10-18
Bugfix for #200

### LINKED ISSUE #200: Step Counter cache problem and moving step of total
# What is not working?

Currently, the step counter is working normally, but there is a slight problem saving the steps after the day. I'm worried about this problem.
 
As far as I know, when i infer the height of the blue square that appeared suddenly, it is confirmed that the current total steps is stored in `lastDoW`.

The `total-steps` is stored in the `lastDoW`.

we have to fix it or **wontfix**?

*I think it's a very minor problem, and I think it's a small exception and it's a part of it that doesn't cause much trouble with the operation.*

# How to reproduce this

1. Running and working , and check step counter

2. Please turn off the timeout and observe the step counter movement in midnight.

# What should happen instead?

# Environment
- **Model** [e.g. Light Edition]
`Light Edition 3.3`
- **OSW-OS** [e.g. Git commit Hash]
-`last commit`
- **Browser** [e.g. Firefox, Chrome]

# Additional context

simonmicro:
Fixes #200

### DIFF
--- a/src/hal/environment.cpp
+++ b/src/hal/environment.cpp
@@ -82,8 +82,8 @@ float OswHal::Environment::getAccelerationZ() {
     return this->accelSensor->getAccelerationZ();
 }
 
-void OswHal::Environment::setupStepStatistics() {
 #ifdef OSW_FEATURE_STATS_STEPS
+void OswHal::Environment::setupStepStatistics() {
     if(!this->accelSensor)
         throw std::runtime_error("No acceleration provider!");
     Preferences prefs;
@@ -154,8 +154,8 @@ void OswHal::Environment::setupStepStatistics() {
     stepHistoryDbgMsg += "}";
     OSW_LOG_D(stepHistoryDbgMsg);
 #endif
-#endif
 }
+#endif
 
 uint32_t OswHal::Environment::getStepsToday() {
     if(!this->accelSensor)
--- a/src/hal/environment.cpp
+++ b/src/hal/environment.cpp
@@ -82,81 +82,6 @@ float OswHal::Environment::getAccelerationZ() {
     return this->accelSensor->getAccelerationZ();
 }
 
-#ifdef OSW_FEATURE_STATS_STEPS
-void OswHal::Environment::setupStepStatistics() {
-    if(!this->accelSensor)
-        throw std::runtime_error("No acceleration provider!");
-    Preferences prefs;
-    bool res = prefs.begin(PREFS_STEPS, false); // Open in RW, just in case
-    assert(res);
-    if(prefs.getBytes(PREFS_STEPS_STATS, &this->_stepsCache, sizeof(this->_stepsCache)) != sizeof(this->_stepsCache)) {
-        // Uoh, the steps history is not initialized -> fill it with zero and do it now!
-        for(size_t i = 0; i < 7; i++)
-            this->_stepsCache[i] = 0;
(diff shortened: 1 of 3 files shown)


RuffaloLavoisier:
CI has reduced time very well. This is the expected result!

src/hal/environment.cpp
```diff
@@ -129,48 +155,39 @@ void OswHal::Environment::setupStepStatistics() {
...
+            if(i == currDoW)
+                stepHistoryDbgMsg += "[";
+            stepHistoryDbgMsg += this->_stepsCache[i];
+            if(i == currDoW)
+                stepHistoryDbgMsg += "]";
+        }
+        stepHistoryDbgMsg += "}";
+        OSW_LOG_D(stepHistoryDbgMsg);
     }
-    stepHistoryDbgMsg += "}";
```
RuffaloLavoisier, line 154: Dbg....? Debug? how about `stepHistoryLog`?
Because abbreviated words and too long. Actually, It doesn't really matter.

src/hal/environment.cpp
```diff
@@ -129,48 +155,39 @@ void OswHal::Environment::setupStepStatistics() {
         if ((resetDay >= 0 && resetDay < 8) && (unsigned short) resetDay == currDoW + 1) // (e.g. 1 - 7 are days, 0 is disabled)
             this->_stepsSum = 0;
 
-        OSW_LOG_D("Updated steps from DoW ", lastDoW, " to DoW ", currDoW);
-        lastDoW = currDoW;
+        this->_stepsLastDoW = currDoW;
+        OSW_LOG_D("Updated steps from DoW ", this->_stepsLastDoW, " to DoW ", currDoW);
```
RuffaloLavoisier, line 159: Doesn't it print with the same value?

RuffaloLavoisier (changes requested):

src/hal/environment.cpp
```diff
@@ -129,48 +155,39 @@ void OswHal::Environment::setupStepStatistics() {
...
         res = prefs.putUInt(PREFS_STEPS_ALL, this->_stepsSum) == sizeof(this->_stepsSum);
         assert(res);
-        res = prefs.putUInt(PREFS_STEPS_DAYLASTCHECKED, lastDoW) == sizeof(lastDoW);
+        res = prefs.putUInt(PREFS_STEPS_DAYLASTCHECKED, this->_stepsLastDoW) == sizeof(this->_stepsLastDoW);
         assert(res);
+        prefs.end();
     }
-    prefs.end();
+
 #ifndef NDEBUG
```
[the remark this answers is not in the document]
simonmicro: Yes, but I do not think it would be bad to give the compiler the clear hint (by removing the code), that he does not even need to construct the string at all...

src/hal/environment.cpp
simonmicro: I mean the correct speaking name would be `stepHistoryDebugMessage`, but I hate that. Your proposal `stepHistoryLog` does not seem to be specific enough and may interfere with names down the road. Sooo... I don't care. If you really think this is bad, YKWTD.

src/hal/environment.cpp
simonmicro: It should, as I've just renamed the variable and moved it into the class itself. This is needed to commit the history only if needed and to prevent reading it every time from the NVS...

RuffaloLavoisier (approved):

PR #317, opened 2022-10-20
CI: only working original upstream repository

RuffaloLavoisier:
Auto formatting working only upstream OSW-OS repository.


simonmicro:
I do not think this is useful - forking users should just disable their own Github Actions instead.

RuffaloLavoisier:
@simonmicro 
Good catch - How can I improve this?

PR #318, opened 2022-10-20
Update connectionInfo method

EliRoberts5:
minor code change

### DIFF
--- a/include/apps/tools/OswAppWebserver.h
+++ b/include/apps/tools/OswAppWebserver.h
@@ -17,6 +17,6 @@ class OswAppWebserver : public OswApp {
 
   private:
     OswUI* ui;
-    void WifiConfigScreen();
+    void ConnectionInfo();
 };
 #endif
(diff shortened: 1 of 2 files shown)


include/apps/tools/OswAppWebserver.h
```diff
@@ -17,5 +17,6 @@ class OswAppWebserver : public OswApp {
 
   private:
     OswUI* ui;
+    void WifiConfigScreen();
```
RuffaloLavoisier, line 20: how about `connectionInfo()`?

RuffaloLavoisier (changes requested):
Thanks for your contribution 👍 

LGTM - But why did you make these changes? It doesn't seem to have changed much.

include/apps/tools/OswAppWebserver.h
EliRoberts5: ok It can be renamed no problem. This change was made to clean up this section with future code work in mind.

EliRoberts5: WifiConfigScreen renamed to ConnectionInfo

RuffaloLavoisier (approved):

PR #320, opened 2022-10-24
Tools screens

EliRoberts5:
Adds a new Tools Screen with 2 pages. New Tools Pages can be added or these default tool pages can be modified or replaced.


src/main.cpp
```diff
@@ -73,10 +76,27 @@
 #define _MAIN_CRASH_SLEEP 2
 #endif
 
+<<<<<<< Updated upstream
 OswAppSwitcher mainAppSwitcher(BUTTON_1, LONG_PRESS, true, true, &main_currentAppIndex);
 OswAppSwitcher watchFaceSwitcher(BUTTON_1, SHORT_PRESS, false, false, &main_watchFaceIndex);
 OswAppSwitcher settingsAppSwitcher(BUTTON_1, SHORT_PRESS, false, false, &main_settingsAppIndex);
 OswAppSwitcher fitnessAppSwitcher(BUTTON_1, SHORT_PRESS, false, false, &main_fitnessAppIndex);
+=======
```
RuffaloLavoisier, line 84: I don't think the files are organized yet.

src/apps/tools/OswAppToolsApp.cpp
```diff
@@ -0,0 +1,26 @@
...
+#include <services/OswServiceTasks.h>
+
+void OswAppToolsApp::setup() {}
+
+void OswAppToolsApp::loop() {
+    OswHal* hal = OswHal::getInstance();
+    hal->gfx()->setTextSize(2);
+
+    hal->gfx()->setTextCursor(40,60);
+    hal->gfx()->print("Tools Page 1");
```
RuffaloLavoisier, line 21: Also, there is nothing in the added Tools.

RuffaloLavoisier:
Also, please set PR to develop-branch based on develop-branch.

RuffaloLavoisier:
If you're ready, please open a new one

src/main.cpp
EliRoberts5: yes, everything is fine with the tools screen update. I'm not sure why there are other commits being added in, they can be ignored. The tools screens are templates so tools can to the tools screen of the watch, like maybe the level screen. So this is just an update to allow the tools screen to available for that.

src/apps/tools/OswAppToolsApp.cpp
EliRoberts5: The desktop git client doesn't seem to be working right, I will try to get a commit to work with git with the develop branch..

PR #322, opened 2022-10-28
Added a tools screen and set the Stopwatch as the first screen of the Tools app screen

EliRoberts5:
This is another example of the Tools Screen. Currently contains the stopwatch as the first screen and two template screens after that called "Page 1" and "Page 2", which can be modified or replaced. ( the plan is to replace them in the near future ). Another option would be to move the "Water Level" and the "Stop Watch" into this Tools Screen instead of having them separate on the main app switcher. This  could easily be done.

### DIFF
--- a/include/apps/tools/OswAppToolsApp.h
+++ b/include/apps/tools/OswAppToolsApp.h
@@ -1,0 +1,19 @@
+#pragma once
+
+#include <osw_hal.h>
+#include <osw_ui.h>
+#include "osw_app.h"
+
+class OswAppToolsApp : public OswApp {
+  public:
+    OswAppToolsApp(void) {
+        ui = OswUI::getInstance();
+    };
+    virtual void setup() override;
+    virtual void loop() override;
+    virtual void stop() override;
+    ~OswAppToolsApp() {};
+
+  private:
+    OswUI* ui;
+};
--- a/include/apps/tools/OswAppToolsPage2App.h
+++ b/include/apps/tools/OswAppToolsPage2App.h
@@ -1,0 +1,19 @@
+#pragma once
+
+#include <osw_hal.h>
+#include <osw_ui.h>
+#include "osw_app.h"
+
+class OswAppToolsPage2App : public OswApp {
+  public:
+    OswAppToolsPage2App(void) {
+        ui = OswUI::getInstance();
+    };
+    virtual void setup() override;
+    virtual void loop() override;
+    virtual void stop() override;
+    ~OswAppToolsPage2App() {};
+
+  private:
+    OswUI* ui;
+};
--- a/src/apps/tools/OswAppToolsApp.cpp
+++ b/src/apps/tools/OswAppToolsApp.cpp
@@ -1,0 +1,26 @@
+#include "./apps/tools/OswAppToolsApp.h"
+
+#include <config.h>
+#include <gfx_util.h>
+#include <osw_app.h>
+#include <osw_config.h>
+#include <osw_config_keys.h>
+#include <osw_hal.h>
+#include <osw_ui.h>
+#include <services/OswServiceTaskWebserver.h>
+#include <services/OswServiceTaskWiFi.h>
+#include <services/OswServiceTasks.h>
+
+void OswAppToolsApp::setup() {}
+
+void OswAppToolsApp::loop() {
+    OswHal* hal = OswHal::getInstance();
+    hal->gfx()->setTextSize(2);
+
+    hal->gfx()->setTextCursor(40,60);
+    hal->gfx()->print("Tools Page 1");
+
+    hal->requestFlush();
+}
+
+void OswAppToolsApp::stop() {}
--- a/src/apps/tools/OswAppToolsPage2App.cpp
+++ b/src/apps/tools/OswAppToolsPage2App.cpp
@@ -1,0 +1,26 @@
+#include "./apps/tools/OswAppToolsPage2App.h"
+
+#include <config.h>
+#include <gfx_util.h>
+#include <osw_app.h>
+#include <osw_config.h>
+#include <osw_config_keys.h>
+#include <osw_hal.h>
+#include <osw_ui.h>
+#include <services/OswServiceTaskWebserver.h>
+#include <services/OswServiceTaskWiFi.h>
+#include <services/OswServiceTasks.h>
+
+void OswAppToolsPage2App::setup() {}
+
+void OswAppToolsPage2App::loop() {
+    OswHal* hal = OswHal::getInstance();
+    hal->gfx()->setTextSize(2);
+
+     hal->gfx()->setTextCursor(40,60);
+    hal->gfx()->print("Tools Page 2");
+    
+    hal->requestFlush();
+}
+
+void OswAppToolsPage2App::stop() {}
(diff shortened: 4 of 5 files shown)


include/apps/tools/OswAppToolsApp.h
```diff
@@ -0,0 +1,19 @@
...
+#include "osw_app.h"
+
+class OswAppToolsApp : public OswApp {
+  public:
+    OswAppToolsApp(void) {
+        ui = OswUI::getInstance();
+    };
+    virtual void setup() override;
+    virtual void loop() override;
+    virtual void stop() override;
```
simonmicro, line 14: Virtual in parent class -> no need to respecify them in an abstract child class.

include/apps/tools/OswAppToolsApp.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+#include <osw_hal.h>
+#include <osw_ui.h>
+#include "osw_app.h"
+
+class OswAppToolsApp : public OswApp {
```
simonmicro, line 7: This is not (really) needed, because... See comment in main file.

include/apps/tools/OswAppToolsPage2App.h
```diff
@@ -0,0 +1,19 @@
...
+        ui = OswUI::getInstance();
+    };
+    virtual void setup() override;
+    virtual void loop() override;
+    virtual void stop() override;
+    ~OswAppToolsPage2App() {};
+
+  private:
+    OswUI* ui;
+};
```
simonmicro, line 19: Same as the other class - additionally, have you heard of `template` classes?

src/apps/tools/OswAppToolsApp.cpp
```diff
@@ -0,0 +1,26 @@
...
+    OswHal* hal = OswHal::getInstance();
+    hal->gfx()->setTextSize(2);
+
+    hal->gfx()->setTextCursor(40,60);
+    hal->gfx()->print("Tools Page 1");
+
+    hal->requestFlush();
+}
+
+void OswAppToolsApp::stop() {}
```
simonmicro, line 26: Dummy code // can be removed.

src/apps/tools/OswAppToolsPage2App.cpp
```diff
@@ -0,0 +1,26 @@
...
+    OswHal* hal = OswHal::getInstance();
+    hal->gfx()->setTextSize(2);
+
+     hal->gfx()->setTextCursor(40,60);
+    hal->gfx()->print("Tools Page 2");
+    
+    hal->requestFlush();
+}
+
+void OswAppToolsPage2App::stop() {}
```
simonmicro, line 26: Same as before.

src/main.cpp
```diff
@@ -73,10 +76,27 @@
...
+uint16_t settingsAppIndex = 0;
+uint16_t fitnessAppIndex = 0;
+uint16_t ToolsAppIndex=0;
+
+OswAppSwitcher mainAppSwitcher(BUTTON_1, LONG_PRESS, true, true, &mainAppIndex);
+OswAppSwitcher watchFaceSwitcher(BUTTON_1, SHORT_PRESS, false, false, &watchFaceIndex);
+OswAppSwitcher settingsAppSwitcher(BUTTON_1, SHORT_PRESS, false, false, &settingsAppIndex);
+OswAppSwitcher fitnessAppSwitcher(BUTTON_1, SHORT_PRESS, false, false, &fitnessAppIndex);
+OswAppSwitcher ToolsAppSwitcher(BUTTON_1, SHORT_PRESS, false, false, &ToolsAppIndex);
+>>>>>>> Stashed changes
```
simonmicro, line 99: Broken code - this are `git` markings, which you'll need to resolve (VSCode should also highlight and warn about this). This completely prevents this from being merged (aside from the other stuff)..

src/main.cpp
```diff
@@ -73,10 +76,27 @@
...
+RTC_DATA_ATTR uint16_t watchFaceIndex;  // Will only be initialized after deep sleep inside the setup() ↓
+uint16_t settingsAppIndex = 0;
+uint16_t fitnessAppIndex = 0;
+uint16_t ToolsAppIndex=0;
+
+OswAppSwitcher mainAppSwitcher(BUTTON_1, LONG_PRESS, true, true, &mainAppIndex);
+OswAppSwitcher watchFaceSwitcher(BUTTON_1, SHORT_PRESS, false, false, &watchFaceIndex);
+OswAppSwitcher settingsAppSwitcher(BUTTON_1, SHORT_PRESS, false, false, &settingsAppIndex);
+OswAppSwitcher fitnessAppSwitcher(BUTTON_1, SHORT_PRESS, false, false, &fitnessAppIndex);
+OswAppSwitcher ToolsAppSwitcher(BUTTON_1, SHORT_PRESS, false, false, &ToolsAppIndex);
```
simonmicro, line 98: Variable names are always _lowercase_.

src/main.cpp
```diff
@@ -199,6 +220,13 @@ void loop() {
         settingsAppSwitcher.paginationEnable();
         mainAppSwitcher.registerApp(&settingsAppSwitcher);
 
+        // new tools screen
+        ToolsAppSwitcher.registerApp(new OswAppToolsApp());
+        ToolsAppSwitcher.registerApp(new OswAppToolsPage2App());
+        ToolsAppSwitcher.paginationEnable();
+        mainAppSwitcher.registerApp(&ToolsAppSwitcher);
```
simonmicro, line 227: I see you've used the concept of the tools screen here (utilizing the app switcher), but I don't really get why you introduced the new classes for the tool "pages". These are dummy classes and their page it intended to be overwritten - right? So why not adapting a similar implementation like the other switchers? Like creating a tools-switcher and then registering other apps into it (instead of into the main switcher). You could also think about adding the concept of "pages" into the app switchers, but this is only an idea...

So, what do you want to do with this? :confused: :thinking:

simonmicro (changes requested):

include/apps/tools/OswAppToolsApp.h
EliRoberts5: yes, I will take a look at that ..

include/apps/tools/OswAppToolsApp.h
EliRoberts5: ok

include/apps/tools/OswAppToolsPage2App.h
EliRoberts5: yes, there is a plan to replace the template code with functional code, for say - various tools.. I will keep working on things and can re-submit something more appropiate. I did move some other apps around and it seemed to be more convenient when using the watch.

src/apps/tools/OswAppToolsApp.cpp
EliRoberts5: yes, I will do that with future updates

src/main.cpp
EliRoberts5: hungarian naming convention.. please clarify what exactly you want changed

include/apps/tools/OswAppToolsApp.h
EliRoberts5: I just copied what was already there .. I will have to look at it

PR #324, opened 2022-11-02
Implement automatic daylight offsets

simonmicro:
- Implemented automatic daylight offset handling
- Minor cleanup and typos
- Added string config key support to emulator
- Switched to POSIX time offset / DST specification string
- Renamed timezones to primary / secondary
- Removed previous DST checker / daylight checkbox
- Added headless mode for the emulator
- Unit-Tests!
  - Serial
  - Emulator
  - Some utilities

- Synchronized time handling across platforms
- Removed ESP device from emulator (moved to virtual) 
- Fixed memory leaks and dangling devices after hard resets


uvwxy:
I think it's great you approach the DST topic. One of the reasons I haven't looked into this is because there are too many variants to consider... and not everyone lives in the EU.

See: https://en.wikipedia.org/wiki/Daylight_saving_time_by_country


There is a lib that covers the US & EU switch: https://github.com/andydoro/DST_RTC/blob/master/DST_RTC.cpp

(which avoids the loop by subtracting the weekday to get to the last Sunday - is this the part you meant about time consuming?)

simonmicro:
@uvwxy Good catch! I think I'll rework this PR to provide a more generic solution to the problem - meaning I plan to implement the table of https://en.wikipedia.org/wiki/Daylight_saving_time_by_country .

PR #330, opened 2022-11-28
CI: scheduling workflow and remove model

RuffaloLavoisier:
- Schedule the workflow to night time (based on UTC).
- CI run currently supported languages and models across all OS.
- Reduce CI time by eliminating unused or redundant models.

⚠️ This only works in the default branch.


.github/workflows/test-OS.yml
```diff
@@ -1,51 +1,37 @@
-name: Build OSW-OS (Ubuntu, Windows, Mac OS, default model)
+name: Build OSW-OS (Ubuntu, Windows, Mac OS, all model)
 
 on:
-  workflow_dispatch:
```
simonmicro, line 4: Hey, I still want to be able to trigger it manually!

.github/workflows/test-OS.yml
```diff
@@ -1,51 +1,37 @@
...
-          skip_after_successful_duplicate: 'true'
-
   Find-feature:
     runs-on: ubuntu-latest
-    needs: check_skip
-    if: ${{ needs.check_skip.outputs.should_skip != 'true' }}
-    steps:         
+    steps: 
+      - name: Check TimeZone
+        run: echo $(date)
```
simonmicro, line 12: I assume this is not needed anymore, right?

platformio.ini
```diff
@@ -41,16 +41,14 @@ extra_scripts =
 build_unflags = -std=gnu++11 # The correct flag will be set by the cppflags python script...
 
 ; Light edition by hardware revisions
-[env:LIGHT_EDITION_V3_2]
```
simonmicro, line 44: Just to be safe: you checked the docs and other documents (like the README) for any mentions?

platformio.ini
```diff
@@ -71,32 +69,4 @@ extra_scripts =
...
-	-D OSW_TARGET_PLATFORM_HEADER='"platform/GPS_EDITION_V3_1.h"'
-	-D PROGMEM_TILES
-	-D GPS_EDITION_ROTATED
-	-D BOARD_HAS_PSRAM
-	-mfix-esp32-psram-cache-issue
-	-D OSW_FEATURE_STATS_STEPS
-	-D OSW_FEATURE_WIFI
-	-D OSW_FEATURE_WIFI_APST
-	-D OSW_FEATURE_WIFI_ONBOOT
-build_type = debug
```
simonmicro, line 102: I would prefer to leave them in, but either comment them or remove them from the pipeline. I would _strongly_ recommend you to port the `setMatrix.sh` script to Python, so I (or anybody else) could easily understand it or - as needed here - patch in a "set of ignored" models, which are not built by the CI anymore (like the GPS* stuff).

simonmicro (changes requested):

simonmicro (approved):

PR #331, opened 2022-11-28
New Release: v0.13.0

### LINKED ISSUE #328: CI: fix warning
Deprecated grammar needs to be changed.

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

RuffaloLavoisier:
- [x] To test #330 
- [x] To fix #328 
- [x] Wait for changes to the workflows #336


.github/buildEverything.py
```diff
@@ -92,6 +94,7 @@ def doBuild(makeDebug):
     ap.add_argument("-l", "--support-language", type=str, required=True, help="# model language to compile. (Enter 'all' to compile all language packs.)")
     ap.add_argument("-m", "--support-model", type=str, required=True, help="# model type to compile. (Enter 'all' to compile all model packs.)")
     ap.add_argument("-f", "--support-feature", type=str, required=False, default="", help="# feature to compile. (Enter a feature to compile.)")
+    ap.add_argument("-b", "--support-build", type=str, required=False, default="", help="# Build configuration to compile. (Enter a Debug/Release to compile.)")
```
simonmicro, line 97: Not required, but no builds will be build without it. We should have a default or require it.

.github/workflows/test-OS.yml
```diff
@@ -0,0 +1,75 @@
...
+    steps:
+    - name: Checkout repository and submodules
+      uses: actions/checkout@v2
+      with:
+        submodules: recursive
+    - name: Cache pip
+      uses: actions/cache@v2
+      with:
+        path: ~/.cache/pip
+        key: cache-${{ secrets.CACHE_VERSION }}-${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
```
simonmicro, line 54: Based on the known cache keys the `hashFiles` does not work... Right?

emulator/src/nvs_flash.cpp
```diff
@@ -13,7 +13,10 @@ bool nvs_flash_erase() {
 }
 
 bool nvs_flash_init() {
-    std::filesystem::create_directories(Preferences::preferencesFolderName);
-    std::cout << "Created NVS path: " << Preferences::preferencesFolderName << std::endl;
+    if(!std::filesystem::exists(Preferences::preferencesFolderName)) {
+        std::filesystem::create_directories(Preferences::preferencesFolderName);
+        std::cout << "Created NVS path: " << Preferences::preferencesFolderName << std::endl;
```
simonmicro, line 18: I think I should migrate the emulators logging to the central system as well...

include/apps/tools/OswAppWebserver.h
```diff
@@ -17,5 +17,6 @@ class OswAppWebserver : public OswApp {
 
   private:
     OswUI* ui;
+    void ConnectionInfo();
```
simonmicro, line 20: Functions should be lowercase - how did that happen?

simonmicro (changes requested):

simonmicro (approved):

.github/buildEverything.py
RuffaloLavoisier: If not given, build debugs and releases at the same time. This is the default I thought of.

.github/buildEverything.py
```diff
@@ -82,8 +82,10 @@ def doBuild(makeDebug):
             exit(2)
         # "Export" firmware.bin
         shutil.copy(os.path.join('.pio', 'build', edition, 'firmware.bin'), os.path.join('.', filename))
-    doBuild(True)
-    doBuild(False)
+    if str(args["support_build"]) == "" or str(args["support_build"]) == "debug":
+        doBuild(True)
+    if str(args["support_build"]) == "" or str(args["support_build"]) == "release":
+        doBuild(False)
```
RuffaloLavoisier, line 88: @simonmicro please check it.

simonmicro: oh, I see - my bad.

PR #332, opened 2022-12-06
Migrate multiple constexpr-s to C++20 syntax

simonmicro:
- Ports template-chains to `if ... ( require {} )`
- Switched to `auto` concepts
- Fixed some more `String`-class bugs

**THIS PR IS NOT WORKING YET; THE ESP-TOOLCHAIN IS TOO OLD / HAS A BUG, WHICH PREVENTS THIS FROM COMPILING!**

_Reference of upstream progress:_ https://github.com/espressif/crosstool-NG/issues/36


simonmicro:
Upstream toolchain is still broken :roll_eyes:

PR #337, opened 2023-01-05
Create the notification system

cutieskye:


simonmicro:
Ohmmm, why did you delete all unit-tests except yours?

include/date.h
```diff
@@ -0,0 +1,8234 @@
+#ifndef DATE_H
+#define DATE_H
```
simonmicro, line 2: Cool lib - but please add it either as a git submodule (available for both emulator and hardware) or as a dependency into platformio (available for hardware).

include/config_defaults.h
```diff
@@ -236,6 +236,10 @@
 #define SERVICE_BLE_COMPANION 0
 #endif
 
+#ifndef SERVICE_NOTIFIER
+#define SERVICE_NOTIFIER 1
+#endif
```
simonmicro, line 241: You are adding a feature to the OS? Please prefix it with `OSW_` accordingly (I know we are not consistent with the old features, but we don't want to break compatibility either).

include/services/NotifierClient.h
```diff
@@ -0,0 +1,30 @@
...
+#include "./services/OswServiceTaskNotifier.h"
+
+#include <string>
+
+class NotifierClient
+{
+public:
+    NotifierClient(const std::string publisher);
+
+    std::pair<std::chrono::time_point<std::chrono::system_clock, std::chrono::seconds>, Notification> createNotification(
```
simonmicro, line 14: I would recommend you to use a `typedef` or to move this into a struct - it makes reading so much easier.

include/services/OswServiceTasks.h
```diff
@@ -3,21 +3,26 @@
...
 class OswServiceTaskExample;
 class OswServiceTaskMemMonitor;
+class OswServiceTaskNotifier;
 #ifdef OSW_FEATURE_WIFI
 class OswServiceTaskWiFi;
 class OswServiceTaskWebserver;
 #endif
-namespace OswServiceAllTasks {
+namespace OswServiceAllTasks
+{
```
simonmicro, line 12: Well, cool that you formatted all that, but I fear our GitHub Actions will revert it to our coding style anyways - so, why do it in the first place?

src/services/NotifierClient.cpp
```diff
@@ -0,0 +1,29 @@
+#include "./services/NotifierClient.h"
+
+NotifierClient::NotifierClient(const std::string publisher) : publisher{publisher}
+{
+}
+
+std::pair<std::chrono::time_point<std::chrono::system_clock, std::chrono::seconds>, Notification> NotifierClient::createNotification(
+    const std::chrono::time_point<std::chrono::system_clock, std::chrono::seconds> timeToFire,
+    const std::string message, const std::array<bool, 7> daysOfWeek)
```
simonmicro, line 9: Warning: With `const std::string message` you are making a copy of the string itself - try to use `const std::string& message` instead! The `const` modifier won't help you here (I think).

src/services/OswServiceTaskNotifier.cpp
```diff
@@ -0,0 +1,130 @@
...
+    auto utcTime = std::chrono::system_clock::from_time_t(OswHal::getInstance()->getUTCTime());
+    auto currentTime = utcTime + std::chrono::seconds{static_cast<int>((OswConfigAllKeys::timeZone.get() + OswConfigAllKeys::daylightOffset.get()) * 3600)};
+    if (auto it = scheduler.begin();
+        it != scheduler.end() && currentTime >= it->first)
+    {
+#ifdef OSW_EMULATOR
+        OSW_LOG_I("Fired a notification");
+        auto t = std::chrono::system_clock::to_time_t(it->first);
+        OSW_LOG_I(std::put_time(std::localtime(&t), "%F %T.\n"));
+        OSW_LOG_I(it->second.getMessage());
```
simonmicro, line 111: I would argue to put this into `OSW_LOG_D` instead

simonmicro (changes requested):

akmal-threepointsix:
> Ohmmm, why did you delete all unit-tests except yours?

Hello, I am the second contributor of this PR.

Were these unit tests necessary? I deleted them because:

- `logging.cpp` didn't compare obtained and expected results (no calls to `EXPECT` or `ASSERT`), so it didn't automatically test anything, just printed some text to console and litteted the actual tests outputs.
- `string.cpp` I deleted them because I was confused by `String` class which inherits from `std:string` (and it's not a good practice to inherit from std classes). But now, as I can see, we need this `String` class for Arduino, am I right? In this case we should return these tests, they seem good and useful.

simonmicro:
>     * `logging.cpp` didn't compare obtained and expected results (no calls to `EXPECT` or `ASSERT`), so it didn't automatically test anything, just printed some text to console and litteted the actual tests outputs.

Yes, it just prints to the console, as I had never any motivation to capture the output and check it - so, yes it is not really testing... Except the fact that it test compile-ability (take a look at the chained templates inside the `OswLogger`-class), running without crashing and (if I look at them) "do they look right". Feel free to friend / inherit the OswLogger class and patch in the output redirect necessary, or open an issue - maybe I'll fix that...

>     * `string.cpp` I deleted them because I was confused by `String` class which inherits from `std:string` (and it's not a good practice to inherit from std classes). But now, as I can see, we need this `String` class for Arduino, am I right? In this case we should return these tests, they seem good and useful.

Indeed - furthermore, the `String` class has a lot of custom (hacky) features to work like the Arduino ones and I broke it more than once by accident. Therefore there are tests - and while inheriting from `std::string` is not so cool, it allows the use of other libraries out there, which are not written for Arduino - take "ArduinoJson", which on the PC-platform only compiles with `std::string` support. Finally, I saved me a lot of work to re-implement another custom String-class by just reusing `std::string` :)

src/services/NotifierClient.cpp
akmal-threepointsix: You are right, for the arguments passed by values `const` modifier is useless. We will remove these unnecessary `const` from now on (we have a few of them in the rest of our code).

However, I would argue that in modern C++ we should always pass by value instead of `const &` (I would only use `const &` for very large objects). Source: https://stackoverflow.com/questions/51705967/advantages-of-pass-by-value-and-stdmove-over-pass-by-reference

simonmicro: Oh - according to https://stackoverflow.com/a/51706522/11664229, this is a valid point. I would still argue as most classes of the OS / apps do not even implement the copy-constructor correctly, just hoping that the default move-constructor works in a more intelligent way is a bad bet. Therefore preventing two copies in the worst case, I would still take the one-copy in every case version. Still, the `std`-libs should work correctly, so feel free to treat this accordingly (I'll try to better myself).

include/services/OswServiceTasks.h
cutieskye: "editor.formatOnSave": true

src/services/OswServiceTaskNotifier.cpp
```diff
@@ -0,0 +1,127 @@
...
+    }
+    auto pair = std::make_pair(timeToFire, notification);
+    scheduler.insert(pair);
+    return pair;
+}
+
+std::vector<NotificationData> OswServiceTaskNotifier::readNotifications(const std::string publisher)
+{
+    const std::lock_guard<std::mutex> lock{mutlimapMutex};
+    std::vector<NotificationData> result{};
```
simonmicro, line 67: Consider changing this to a list, as you do not need to waste reallocation cycles here :)

simonmicro:
Regarding the missing unit-test checks for the logger: There you go (I've just had some spare time) -> https://github.com/Open-Smartwatch/open-smartwatch-os/pull/341

src/services/OswServiceTaskNotifier.cpp
cutieskye: I'd still use a vector as the default container. We only need to increase the vector's capacity once, as we know it won't ever hold more notifications than the multimap has.

simonmicro: Oh, fixed in 99f6bf4a912f24118bee9159f61df921c90a8094 -> ignore me

simonmicro (approved):

PR #343, opened 2023-01-08
Feature/timer

akmal-threepointsix:
- I added timer feature
- It is completely working, i.e. you can set timer, pause/resume, reset, or simply wait till it ends
- When it ends, it uses notification system (`NotificationClient` and `OswServiceTaskNotifier`) to notify the user

Some screenshots of the timer:

1. Initial screen
image
2. Set timer screen
image
3. Running timer screen
image


include/locales/en-US.h
```diff
@@ -12,6 +12,12 @@
 #define LANG_CANCEL "Cancel"
 #define LANG_SAVE "Save"
 
+// App: Timer
+#define LANG_TIMER "Timer"
+#define LANG_START "Start"
+#define LANG_PAUSE "Pause"
+#define LANG_RESUME "Resume"
+
```
cutieskye, line 20: We need to devise a way for an application to fall back to the English language if it has not been localized yet.

cutieskye (changes requested):

include/locales/en-US.h
akmal-threepointsix: I think your idea is good, but it should be opened as a separate issue and implemented in a separate branch.
This PR only implements the Timer feature.

akmal-threepointsix: Fixed the error

include/apps/clock/OswAppTimer.h
```diff
@@ -0,0 +1,60 @@
...
+    // Helpers
+    void timestampToSec();
+    void resetTimer();
+
+    // Draw on screen
+    void drawNumber(const int number, const int index);
+    void drawTime(const int totalSeconds);
+
+    OswAppSwitcher *clockAppSwitcher{};
+    NotifierClient notifierClient{"com.akmal-threepointsix.application.timer"};
```
simonmicro, line 47: IIRC you plan to merge this into OSW itself -> maybe use `org.open-smartwatch.osw.timer` instead?

include/config_defaults.h
```diff
@@ -217,15 +217,15 @@
 #ifndef TOOL_CALCULATOR
 #define TOOL_CALCULATOR 1
 #endif
-#ifndef TOOL_STOPWATCH
-#define TOOL_STOPWATCH 1
-#endif
```
simonmicro, line 222: Why was this removed? Does your app replace the timer? No. Did you replace the flag? No.

Also you missed one in the `CMakeLists.txt` for the emulator.

Maybe you could introduce a new flag to make the whole "clock-apps" section optional.

simonmicro (changes requested):

simonmicro (approved):

PR #347, opened 2023-01-15
Show notifications at appropriate times

cutieskye:
This pull request integrates my notification system and feature/UInotifications by @simonmicro.


simonmicro (approved):
Sorry, could you either follow the OSW-code-style (see workflow files in `.github`) or just disable your format-on-save? It is no fun to review such mess (and won't make it faster).

PR #358, opened 2023-04-01
Tutorial, Icons, better rendering & more

### LINKED ISSUE #334: OswAppDrawer (better app-switcher UI)
We noted several problems with the current UI:

# Problems
- Apps tend to use all three buttons - especially the currently needed button-1 long-/short-press -> reduce it to just button-1 long-press
- Many apps? Horrible. You have to scroll through to get to the right one - which IS AWFULLY SLOW! -> a select-your-app ui?
- No overview about which apps are available -> a select-your-app ui?

# Solution!
- [x] app-drawer (replaces switcher with interactive, categorized menu of apps with icons)
  - [x] late-init of apps for app-drawer
     - [x] container, with `get()`-able pointer to instance
     - [x] apps constructed only created as needed
     - [x] cleaned-up upon app section
     - [x] has (OswAppV2*)() pointers to init functions for app-instances
  - [x] sleep-persistance -> store "index" of last selected app and recover from deep-sleep to that app
- [x] ~allow override of getButtonStateMask, so the drawer can pass through the app and patch-in itself~ -> solved by using the apps own button logic as wrapper
- [x] remove old switcher

# Nice-to-have
- [x] ~animations!~ -> got the classic ones, but I plan not to add many more for now...

### LINKED ISSUE #353: Tutorial Application
# Screens
- [x] welcome user
- [x] let user press different button press-modes
- [x] show note about battery calibration
- [x] instruct how to open app drawer
- [x] show known hardware-problems:
  - [x] no battery level with wifi
  - [x] wifi causes cpu brown-out
  - [x] on gps-edition: reception is terrible
- [x] MUST self-timeout after 60 seconds per step (because e.g. some hardware problems could user prevent from completion)

# More
- [x] ~remove example application; replace it with the tutorial app, which uses the `OswAppV2` interface~
- [x] also (re)add oswappv1 example with same features and show how to use it
- [x] add ref to example app also into readme / docs
- [x] shown if build commit-hash changes
- [x] after tutorial the wrong app is shown?

### LINKED ISSUE #354: OswAppV2 (new osw app class)
The primary motivation is to only render the screen when needed (needed to port to Watchy / E-Ink displays). Turns out, when we start providing more features in the app-class (like callbacks, metadata, "events") we can even add more unified features (button-handling, animations, overlay-menus)!

Here a simple draft of the new app-class:
```c++
class OswHal;
class OswAppV2 {
  public:
    virtual const char* getAppId() = 0;
    virtual const char* getAppName() = 0;
    virtual ???& getAppIcon() = 0;

    // TODO all those ↓ must update hal-reference
    virtual void onStart();
    virtual void onLoop();
    virtual void onDraw();
    virtual void onStop();

    virtual void onButton(void id, bool down, bool shortPress, bool longPress);
#ifdef OSW_EMULATOR
    virtual void onLoopDebug() {}; // By default no debug loop (GUI) is implemented
#endif

  protected:
    OswHal* hal = nullptr; // You guys are needing that anyways (but you often cache incorrectly), so it is now given to you <3
    bool needsRedraw = false;
};
```

More stuff:
- [x] implement base class
- [x] flags (keep-awake, no overlay / force-overlay)
- [x] compatibility for app v2 to wrap app v1
- [x] think about @RuffaloLavoisier idea for multi-button menus (and their integration here) -> they should be part of the tutorial anyways
- [x] port / wrap legacy (v1) apps
  - [x] design icons? code-based icons vs `PROGMEM` icons?
  - [x] setting the default watchface does not seem to work on the real hardware -> the event `Button event for app osw.wf.fit: id? 2 up? 1 state? 2` is missing - only `Button event for app osw.wf.fit: id? 2 up? 0 state? 2` is being reported.
- [x] `OswAppV2Compat` allow for keeping the screen on (by default?) -> set that also properly inside the `main.cpp` for the old apps

### LINKED ISSUE #355: OswIcons, Button-Aliases and more small stuff
Here is a list of random thoughts...

# Buttons
- [x] button-aliases -> button 1/2/3 will be button_select, button_up, button_down -> finally using an enum
- [x] (extra-)long-press animations on all buttons
- [x] unify button-modes -> click, double click, long press, very long press

# App-Drawer
- [x] osw-icons
  - [x] Python script to take any 16x16 or 32x32 png to convert into monochromatic icon-bytes
  - [x] class to allow simple rendering at coords x, y
  - [x] should support coloring options during draw
  - [x] allow for converting of classic images (then not as oswicon, but as oswimage (?))
- [x] Implement #334 
- [x] If capability is not given on button, do not even show the animation for it (like no long-press-bit -> no animation during long press)

# Stuff
- [x] abstract osw config key category to merge names + be represented in own class -> EXCEPT if compiler merges them anyways
- [x] ~use app-id -> notifications?, wakeup-configs?, ...more?~
- [x] osw-os ui class.activeApp member (if we don't already have that one)

simonmicro:
Closes #334
Closes #355
Closes #354
Closes #353

Pending TODOs:
- [x] translate new strings
- [x] fix all newly introduced `TODO` markings
- [x] port / wrap all current osw-apps - maybe @RuffaloLavoisier helps?
- [x] make `OswIconProgmem` `const static` to reduce copies during drawing with changed properties


simonmicro:
At all reviewers: Feel free to also try the new parts on live hardware and try to find bugs! Sometimes specific timings / usage patterns only occur on the real hardware, may then leading to bugs.

RuffaloLavoisier (approved):

PR #351, opened 2023-02-15
Testing/integration tests

akmal-threepointsix:
Main changes:

- Added ImGUI Test Engine (`CMakeLists.txt`, `emulator/lib/ImGUI_TestEngine`)
- Updated ImGUI, because older version was incompatible with ImGUI Test Engine (`emulator/lib/ImGUI`)
- Added UI tests for Alarm and Timer (`emulator/src/UiTests/OswAppAlarm.cpp`, `emulator/src/UiTests/OswAppTimer.cpp`)
- Updated `README.md` with new testing chapter
- I didn't change any production code, because testing must not interfere with business logic. The only exception is that I added friend class in `include/apps/main/switcher.h` to access private fields of `OswAppSwitcher`.

Finally, I know you want to see how the tests look like. Here is the demo in max speed:

https://user-images.githubusercontent.com/30224608/219116887-a7c91519-42e1-4795-b345-2ec11aa44a2a.mp4

And here is the normal speed:

https://user-images.githubusercontent.com/30224608/219117057-7f0c12be-1816-41ab-a263-5da0e1b07e02.mp4

Closes #340

PR #367, opened 2023-07-09
Testing/integration tests

akmal-threepointsix:
Since this old [PR #351](https://github.com/Open-Smartwatch/open-smartwatch-os/pull/351) diverged too much from the up-to-date develop branch, I decided to create a new branch from the develop and open this new PR.

These changes are basically the same as in [PR #351](https://github.com/Open-Smartwatch/open-smartwatch-os/pull/351):

- Added ImGUI Test Engine submodule (`CMakeLists.txt`, `emulator/lib/ImGUI_TestEngine`). **IMPORTANT:** since I added new submodule, you need to run `$ git submodule update --init --recursive` after checkout to this branch
- Updated ImGUI, because older version was incompatible with ImGUI Test Engine (`emulator/lib/ImGUI`)
- Added UI tests for Alarm and Timer (`emulator/src/UiTests/OswAppAlarm.cpp`, `emulator/src/UiTests/OswAppTimer.cpp`)
- Updated `README.md` with new testing chapter
- I didn't change any production code, because testing must not interfere with business logic. I only added three friend classes in three header files (`emulator/include/Emulator.hpp`, `include/apps/OswAppDrawer.h` and `include/apps/OswAppV2Compat.h`)

And some *NEW* changes:

- I fixed all errors related to the newest [PR #358](https://github.com/Open-Smartwatch/open-smartwatch-os/pull/358). It made my UI testing code a bit uglier (e.g. I had to use a lot of casting), but at the time I am writing this, everything at least works...
- In `UiTest_main.hpp`, I added some code to skip tutorial, if emulator was launched with `--ui_tests` option


simonmicro:
Hi,
thanks for your work - again! So two things I've noted here...

1. Meh - _le pipeline is failing_...
2. Could you reorder the code to bundle all the unit/ui/helpers stuff for testing under `src/tests` again? You've moved them out and put them into the "regular" code. Do not worry, if you won't do it I'll clean that up :wink: 

After you put so much work into it to re-submit this PR I'll absolutely make sure to get it merged this time. This is the next thing going in - so I'll not allow this to run out-of-sync again! Hehe...

akmal-threepointsix:
> 1. Meh - _le pipeline is failing_...

Fixed (hopefully) by pushing ImGui Test Engine submodule folder

> 2. Could you reorder the code to bundle all the unit/ui/helpers stuff for testing under `src/tests` again? You've moved them out and put them into the "regular" code. Do not worry, if you won't do it I'll clean that up 😉

Done. I forgot about this one :)

akmal-threepointsix:
Ok, I also removed `reinterpret_cast` from `tests/helpers/TestDrawer.h`. 

It turned out, that `OswAppV1` and `OswApp` are the same class, so I can simply `static_cast` timer and alarm from `OswAppV1` to `OswAppTimer` and `OswAppAlarm` respectively

simonmicro:
@akmal-threepointsix If you formulate that like this... Ever thought about the more safe variant of https://en.cppreference.com/w/cpp/language/dynamic_cast ? This also ensures you won't cast in the future if we break it :wink:

akmal-threepointsix:
Yes, sure, `dynamic_cast` is also a valid option and safer one. However, it introduces some run-time overhead, which in this case seemed unnecessary to me. On the other hand, you are right, in future someone might change codebase and `static_cast` will silently fail, so it would be better to use `dynamic_cast`. I believe, run-time cost is insignificant, especially since it is in emulator & UI tests environment, not in production.

akmal-threepointsix:
> @akmal-threepointsix If you formulate that like this... Ever thought about the more safe variant of https://en.cppreference.com/w/cpp/language/dynamic_cast ? This also ensures you won't cast in the future if we break it 😉

Ok, done. I believe you can run workflows, I am curious if it will fail or not this time :)

akmal-threepointsix:
> @akmal-threepointsix If you formulate that like this... Ever thought about the more safe variant of https://en.cppreference.com/w/cpp/language/dynamic_cast ? This also ensures you won't cast in the future if we break it 😉

Ok, done. Changed `static_cast` to `dynamic_cast` and added `assert` to confirm that casting was successful and didn't return `nullptr`

PR #372, opened 2023-11-09
OswAppStepStats: add debug window for current position value

RuffaloLavoisier:
Add a debug window to 'OswAppStepStats' as well.

### DIFF
--- a/src/apps/tools/OswAppStepStats.cpp
+++ b/src/apps/tools/OswAppStepStats.cpp
@@ -85,5 +85,15 @@ void OswAppStepStats::loop() {
     showStickChart();
 }
 
+#ifdef OSW_EMULATOR
+#include "imgui.h"
+
+void OswAppStepStats::loopDebug() {
+    ImGui::Begin("Debug: OswAppStepStats");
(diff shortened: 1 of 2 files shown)


src/apps/tools/OswAppStepStats.cpp
```diff
@@ -85,5 +85,15 @@ void OswAppStepStats::loop() {
     showStickChart();
 }
 
+#ifdef OSW_EMULATOR
+#include "imgui.h"
+
+void OswAppStepStats::loopDebug() {
+    ImGui::Begin("Debug: OswAppStepStats");
+    ImGui::InputScalar("cursorPos", ImGuiDataType_U8, &this->cursorPos);
```
RuffaloLavoisier, line 93: @simonmicro what is mean data type ? (`ImGuiDataType_U8`,,,sometime..`ImGuiDataType_S8`)

simonmicro: This is the datatype ImGui will choose as valid input range via the GUI. It has to match the underlying data size and is used to prevent invalid inputs.

U8 = unsigned, 8-bit
S8 = signed, 8-bit

simonmicro (approved):

PR #376, opened 2023-12-16
 OSW Weather emulator support

Lorenzosciacca:
This pull request introduces changes to enable emulator support in the application OSW Weather. The key enhancement is the ability of the emulator to load a JSON file containing a valid API response from OpenWeather.

Changes:

Added emulator support to OSW Weather
Documentation update

PR #382, opened 2024-04-06
Updated PR #376

simonmicro:
This PR is based on https://github.com/Open-Smartwatch/open-smartwatch-os/pull/376 , but the author did not authorize modifications on its fork before merge - so I forked his branch and pulled it into this PR.

### DIFF
--- a/src/apps/_experiments/OswAppWeather.cpp
+++ b/src/apps/_experiments/OswAppWeather.cpp
@@ -250,7 +250,7 @@ bool OswAppWeather::_request() {
     http.end();
     OswServiceAllTasks::wifi.disconnectWiFi();
     OSW_LOG_D("Request returned code: ", code);
-    if (code != 200) {
+    if (code != HTTP_CODE_OK) {
         OSW_LOG_E("Unexpected API response: ", code);
         this->dataLoaded = false;
         return false;
--- a/src/apps/_experiments/OswAppWeather.cpp
+++ b/src/apps/_experiments/OswAppWeather.cpp
@@ -3,16 +3,21 @@
 #include "apps/_experiments/OswAppWeatherEncoder.h"
 #include "services/OswServiceTaskWiFi.h"
 #include <services/OswServiceTasks.h>
-#include <HTTPClient.h>
 #include <cstring>
 #include <gfx_util.h>
 #include <OswAppV1.h>
 #include <osw_hal.h>
 #include "fonts/DS_DIGI12pt7b.h"
 #include "ArduinoJson.h"
-
+#ifndef OSW_EMULATOR 
+#include <HTTPClient.h>
+#endif
 #define OPENWEATHERMAP_URL "https://api.openweathermap.org/data/2.5/forecast?q="
 #define URL_REQ OPENWEATHERMAP_URL OPENWEATHERMAP_CITY "," OPENWEATHERMAP_STATE_CODE "&appid=" OPENWEATHERMAP_APIKEY "&cnt=24"
+#ifdef    OSW_EMULATOR
+#include <iostream>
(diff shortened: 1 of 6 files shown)


src/apps/_experiments/OswAppWeather.cpp
```diff
@@ -263,21 +279,22 @@ void OswAppWeather::weatherRequest() {
...
+        OSW_LOG_D("free heap ", ESP.getFreeHeap());
         code = http.GET();
     } else {
         return false;
     }
     http.end();
     OswServiceAllTasks::wifi.disconnectWiFi();
-    OSW_LOG_I("code:", code);
+    OSW_LOG_D("Request returned code: ", code);
     if (code == 200) {
```
RuffaloLavoisier, line 298: Can we use replace word? 

```
SUCCESS = 200
{ . . . }
if (code == SUCCESS)
```

src/apps/_experiments/OswAppWeather.cpp
```diff
@@ -288,19 +305,19 @@ bool OswAppWeather::_request() {
         strcpy(encoded_arr, encoded.c_str());
         String encoded_S = String(encoded_arr);
         if (!this->pref.putString("wtr",encoded_S)) {
-            OSW_LOG_E("Error: unable to write to NVS");
+            OSW_LOG_E("Unable to write to NVS?!");
```
RuffaloLavoisier, line 308: what is mean last word `?!`

src/apps/_experiments/OswAppWeather.cpp
```diff
@@ -309,6 +326,18 @@ bool OswAppWeather::_request() {
...
+#else
+void OswAppWeather::weatherRequest() {
+    this->requestMode = true;
+}
+bool OswAppWeather::_request() {
+    // in emulator instantly "fullfill" the request
+    this->requestMode=false;
+    this->dataLoaded=true;
+    return true;
+}
```
RuffaloLavoisier, line 338: we need code-style
```
line 331: this->requestMode = true;
line 335: this->requestMode=false;
```
not important, but i think we need..

src/apps/_experiments/OswAppWeather.cpp
```diff
@@ -379,9 +408,31 @@ void OswAppWeather::printDate() {
 }
 
 bool OswAppWeather::loadData() {
+#ifdef OSW_EMULATOR
+    std::ifstream inFile;
+    inFile.open("file_weather.json"); //open the input file
```
RuffaloLavoisier, line 413: where is this file? local or device?

RuffaloLavoisier (commented):
LGTM

src/apps/_experiments/OswAppWeather.cpp
simonmicro: Hmm, maybe the HTTP lib already has defines for that, lemme check

simonmicro: Bingo: `HTTP_CODE_OK`

src/apps/_experiments/OswAppWeather.cpp
simonmicro: Like a surprised / confused (`?`), but shouting (`!`) - I chose confused, because the error-handling is too bad that the message cannot tell why

src/apps/_experiments/OswAppWeather.cpp
simonmicro: I extended the error message below that - but yeah, how to obtain is still a good question

simonmicro:
Nope, found multiple SEGFAULT-s in the emulator due to this app. I'll patch these first.

PR #383, opened 2024-04-06
Added serial console

simonmicro:
Some simple stuff to configure/reset/reboot/wipe the device without the need to setup/connect to the hotspot.

Also very useful if you disabled AutoAP, messed up the WiFi configuration and are now locked out. ALSO, much more convenient to reconfigure on-the-fly instead of recompiling all the time.

### DIFF
--- a/.github/workflows/test-EMULATOR.yml
+++ b/.github/workflows/test-EMULATOR.yml
@@ -31,7 +31,7 @@ jobs:
       uses: nschloe/action-cached-lfs-checkout@v1.2.1
       with:
         submodules: recursive
-        lfs: 'true'
+        # lfs: 'true'
     - name: Update packages
       run: sudo apt-get update
     - name: Install packages
--- a/.github/workflows/test-EMULATOR.yml
+++ b/.github/workflows/test-EMULATOR.yml
@@ -4,7 +4,7 @@ on:
   push:
     paths_ignore: '["**/*.md", "**/scripts/screen_capture/**"]'
   pull_request:
-    branches: [ master, develop ]
+    branches: [master, develop]
 
 jobs:
   check_skip:
@@ -15,8 +15,8 @@ jobs:
       - id: skip_check
         uses: fkirc/skip-duplicate-actions@v5.2.0
         with:
-          concurrent_skipping: 'same_content_newer'
-          skip_after_successful_duplicate: 'true'
+          concurrent_skipping: "same_content_newer"
+          skip_after_successful_duplicate: "true"
 
   build-EMULATOR:
     runs-on: ubuntu-22.04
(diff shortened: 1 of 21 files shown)


.github/workflows/test-EMULATOR.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       uses: nschloe/action-cached-lfs-checkout@v1.2.1
       with:
         submodules: recursive
-        lfs: 'true'
+        # lfs: 'true'
```
RuffaloLavoisier, line 34: this is dead code? so please remove.

.github/workflows/test-FEATURE.yml
```diff
@@ -28,7 +28,7 @@ jobs:
         uses: nschloe/action-cached-lfs-checkout@v1.2.1
         with:
           submodules: recursive
-          lfs: 'true'
+          # lfs: 'true'
```
RuffaloLavoisier, line 31: also dead code

.github/workflows/test-FEATURE.yml
```diff
@@ -57,7 +57,7 @@ jobs:
       uses: nschloe/action-cached-lfs-checkout@v1.2.1
       with:
         submodules: recursive
-        lfs: 'true'
+        # lfs: 'true'
```
RuffaloLavoisier, line 60: also dead code

.github/workflows/test-OS.yml
```diff
@@ -15,7 +15,7 @@ jobs:
         uses: nschloe/action-cached-lfs-checkout@v1.2.1
         with:
           submodules: recursive
-          lfs: 'true'
+          # lfs: 'true'
```
RuffaloLavoisier, line 18: also dead code

.github/workflows/test-OS.yml
```diff
@@ -50,7 +50,7 @@ jobs:
       uses: nschloe/action-cached-lfs-checkout@v1.2.1
       with:
         submodules: recursive
-        lfs: 'true'
+        # lfs: 'true'
```
RuffaloLavoisier, line 53: also dead code

.github/workflows/test-OSW.yml
```diff
@@ -29,7 +29,7 @@ jobs:
         uses: nschloe/action-cached-lfs-checkout@v1.2.1
         with:
           submodules: recursive
-          lfs: 'true'
+          # lfs: 'true'
```
RuffaloLavoisier, line 32: also dead code

.github/workflows/test-OSW.yml
```diff
@@ -63,7 +63,7 @@ jobs:
       uses: nschloe/action-cached-lfs-checkout@v1.2.1
       with:
         submodules: recursive
-        lfs: 'true'
+        # lfs: 'true'
```
RuffaloLavoisier, line 66: also dead code

src/services/OswServiceTaskConsole.cpp
```diff
@@ -0,0 +1,178 @@
...
+            }
+            Serial.print(c);
+        } else if(c == 9) { // tab
+            Serial.println(); // finish the prompt line
+            this->showHelp();
+            this->showPrompt();
+        } else {
+            Serial.print((char) 0x07); // bell
+            OSW_LOG_D("Unprocessable character (",(int) c, "): ", c);
+        }
```
RuffaloLavoisier, line 35: how about switch-case?
and IMHO, unless it's 1 and 0, all hard-coded constants would be read is difficult better.. if they had an alternative preprocessing variable. Even if you add an annotation comment `// blah blah`..
```
TAB = 9
{ . . . }
c == TAB
```

RuffaloLavoisier (changes requested):
See Review 👍

.github/workflows/test-EMULATOR.yml
simonmicro: I would like to let it stay, as the line above (`# uses: actions/checkout@v3 # incompat`) is also there for documenting that we are now using another action. The `lfs` is strictly required for the commented variant. Do you think we should still remove it?

src/services/OswServiceTaskConsole.cpp
simonmicro: Hmm, the first case cannot be done via a switch case, as it covers a range. The remaining yes, while I still feel this will bloat up the code, it can be done (especially thinking of potential optimizations). I'll include it. :+1:

simonmicro: Done

.github/workflows/test-EMULATOR.yml
RuffaloLavoisier: @simonmicro OOOOH It's not a dead code!!!!, I want to leave `# blah blah`comments for documentation in a good format.. in a better format so that it doesn't look like a dead code...

RuffaloLavoisier (approved):

PR #394, opened 2024-05-21
Add Newline at end of file

zwim:
Reasons see in https://gist.github.com/OleksiyRudenko/d51388345ea55767b7672307fe35adf3

Or more generic
Google .> "why do we want an newline at the end of a file"


src/animations/demos/rotation/leafs_128_32_4x.png
simonmicro, line 1: Hmm, I do not think that this file should have been changed?

src/animations/demos/rotation/watchface01.png
simonmicro, line 1: same here

simonmicro (changes requested):
The images should not change. While PNG may accept these changes, I do not think that this should be done.

simonmicro (approved):

PR #397, opened 2024-06-01
Workflows: release tag and upload assets

RuffaloLavoisier:
image

This workflows,  just ONLY CREATE RELEASE AND UPLOAD ASSETS.


.github/workflows/release.yml
```diff
@@ -0,0 +1,93 @@
...
+      - name: Compile language ${{ matrix.language }} model ${{ matrix.model }}
+        run: python3 .github/buildFirmware.py -l "${{ matrix.language }}" -m "${{ matrix.model }}" -b "${{ matrix.build-configuration }}"
+      - name: DEBUG LS-cmd
+        run: ls -lahR
+      - name: Release
+        uses: softprops/action-gh-release@v2
+        with:
+          files: "*.bin" # TEMP use zip to create an archive again
+          draft: false
+          prerelease: true
```
RuffaloLavoisier, line 92: I will only recommend drafts because the pre-release is posting, and the draft only saves. Is that correct?

simonmicro: I think drafts are indeed good, but I'll also recommend setting pre-release also to true by default.

.github/workflows/release.yml
```diff
@@ -0,0 +1,93 @@
+name: Release OSW-OS Firmware
```
simonmicro, line 1: Maybe we can merge this new workflow into the existing one, which also builds all firmware variations already?

.github/workflows/release.yml
```diff
@@ -0,0 +1,93 @@
...
+        with:
+          python-version: "3.10"
+      - name: Install PlatformIO
+        run: python -m pip install --upgrade pip && pip install --upgrade platformio
+      - name: Rename config
+        run: mv include/config.h.example include/config.h
+      - name: Compile language ${{ matrix.language }} model ${{ matrix.model }}
+        run: python3 .github/buildFirmware.py -l "${{ matrix.language }}" -m "${{ matrix.model }}" -b "${{ matrix.build-configuration }}"
+      - name: DEBUG LS-cmd
+        run: ls -lahR
```
simonmicro, line 86: Please remove this

simonmicro (approved):

PR #400, opened 2024-09-14
Refactoring code chunk

RuffaloLavoisier:
with #399

### DIFF
--- a/src/apps/tools/OswAppDistStats.cpp
+++ b/src/apps/tools/OswAppDistStats.cpp
@@ -9,6 +9,8 @@
 #include <osw_hal.h>
 #include <osw_pins.h>
 
+#define WEEK 7
+
 void OswAppDistStats::drawChart() {
     OswHal* hal = OswHal::getInstance();
     uint8_t chartStickHeight = 55;
@@ -19,7 +21,7 @@ void OswAppDistStats::drawChart() {
     uint32_t dayOfMonth = 0;
     hal->getLocalDate(&dayOfMonth, &weekDay);
 
-    for (uint8_t index = 0; index < 7; index++) {
+    for (uint8_t index = 0; index < WEEK; index++) {
         uint32_t weekDayDist = OswAppWatchfaceFitness::calculateDistance(hal->environment()->getStepsOnDay(index));
         uint16_t chartStickValue = ((float)(weekDayDist > goalValue ? goalValue : weekDayDist) / goalValue) * chartStickHeight;
 
--- a/src/apps/tools/OswAppDistStats.cpp
+++ b/src/apps/tools/OswAppDistStats.cpp
@@ -9,8 +9,6 @@
 #include <osw_hal.h>
 #include <osw_pins.h>
 
-#define WEEK 7
-
 void OswAppDistStats::drawChart() {
     OswHal* hal = OswHal::getInstance();
     uint8_t chartStickHeight = 55;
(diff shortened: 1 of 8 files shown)


src/apps/tools/OswAppDistStats.cpp
```diff
@@ -9,6 +9,8 @@
 #include <osw_hal.h>
 #include <osw_pins.h>
 
+#define WEEK 7
+
```
simonmicro, line 13: I do not think define-ing the number of week days is really that useful - especially as this define is never reuzsed and just re-defined in each source file. If you really want to label it (and I stretch: I you _really_ want to), maybe declare a `constexpr` int before the `for` call and avoid using the pre-processor for this.

src/apps/watchfaces/OswAppWatchface.cpp
```diff
@@ -25,33 +27,34 @@ const char* OswAppWatchface::getAppName() {
 }
 
 #ifdef OSW_FEATURE_STATS_STEPS
-void OswAppWatchface::drawStepHistory(OswUI* ui, uint8_t x, uint8_t y, uint8_t w, uint8_t h, uint32_t max) {
+void OswAppWatchface::drawStepHistory(OswUI* ui, uint8_t posX, uint8_t posY, uint8_t width, uint8_t height, uint32_t max) {
     OswHal* hal = OswHal::getInstance();
     OswUI::getInstance()->resetTextColors();
     uint32_t weekDay = 0;
     uint32_t dayOfMonth = 0;
+    uint8_t index = 0;
```
simonmicro, line 35: Why did you move this out of the `for`-loop if you are not accessing it outside the loop? Just leave it in there to reduce the potential of variable-name collisions.

src/apps/watchfaces/OswAppWatchface.cpp
```diff
@@ -25,33 +27,34 @@ const char* OswAppWatchface::getAppName() {
...
         hal->gfx()->setTextCenterAligned();  // horiz.
         hal->gfx()->setTextBottomAligned();
         hal->gfx()->setTextSize(1);
-        hal->gfx()->setTextCursor(CENTER_X, y - 1);
+        hal->gfx()->setTextCursor(CENTER_X, posY - 1);
 
         hal->gfx()->print(hal->environment()->getStepsToday() + (OswConfigAllKeys::settingDisplayStepsGoal.get() ? String("/") + max:""));
 
-        hal->gfx()->setTextCursor(CENTER_X, y + 1 + 8 + w * 4);
+        hal->gfx()->setTextCursor(CENTER_X, 221);
```
simonmicro, line 57: I would not hard-code such value! The display size is dynamic and the placement of the boxes is too. If you want you could try to tell the compiler to pre-calculate it using a `constexpr`, but this math is really not a slowdown.

src/apps/watchfaces/OswAppWatchfaceMix.cpp
```diff
@@ -24,8 +24,8 @@ const char* OswAppWatchfaceMix::getAppName() {
 void OswAppWatchfaceMix::analogWatchDisplay() {
     OswHal* hal = OswHal::getInstance();
     uint32_t second = 0;
-    uint32_t minute = 0;  // Unused, but required by function signature
-    uint32_t hour = 0;    // Unused, but required by function signature
+    uint32_t minute = 0;
+    uint32_t hour = 0;
 
     hal->getLocalTime(&hour, &minute, &second);
```
simonmicro, line 30: Removed it before, but here... Not?

simonmicro (changes requested):
Some good changes, some... Why?

RuffaloLavoisier:
LGTM.

RuffaloLavoisier:
to #405.

PR #401, opened 2024-09-16
Resize param data type

RuffaloLavoisier:
Hi there ✋ 

PR description.
- `day` max value is `31` (need to 8 bit) and `month` max value is `12` (need to 4 bit), and `weekday` is `7`.. so need to 4 bit.

### DIFF
--- a/emulator/include/RtcDS3231.h
+++ b/emulator/include/RtcDS3231.h
@@ -15,13 +15,13 @@ class RtcDateTime {
 
     void InitWithUnix32Time(time_t time);
 
-    uint32_t Hour();
-    uint32_t Minute();
-    uint32_t Second();
+    uint8_t Hour();
+    uint8_t Minute();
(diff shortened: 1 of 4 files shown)


RuffaloLavoisier:
@simonmicro 
There are many over-declared data types than the data max-value range. What do you think about this?

simonmicro (changes requested):
Really good idea - but you may need to fix a lot of stuff

RuffaloLavoisier:
Why don't we revise the rest of the parts later.

simonmicro (changes requested):
Stuff I found, which should also be changed with this:

- src/apps/watchfaces/OswAppWatchfaceFitnessAnalog.cpp -> 218, 220
- src/apps/watchfaces/OswAppWatchfaceDigital.cpp -> 94

simonmicro (approved):

PR #403, opened 2024-09-28
Fix weekDay API spec

RuffaloLavoisier:
Add OUT param for weekDay char type, remove return type.


RuffaloLavoisier:
Emulator is working.

PR #405, opened 2024-09-29
Simplifies API usage

RuffaloLavoisier:
Update getDate, getTime API

How to use
```
OSW_TIME oswTime = { 0, };
hal->getLocalTime(&oswTime);
```
Each element can be referred to as a sub-element. It returns data for a time or date and can be used by selecting the necessary elements.

### DIFF
--- a/include/osw_hal.h
+++ b/include/osw_hal.h
@@ -28,6 +28,21 @@
 #define ERR_SD_MISSING 1
 #define ERR_SD_MOUNT_FAILED 2
 
+typedef struct {
+    uint32_t hour;
+    uint32_t minute;
+    uint32_t second;
+    bool afterNoon;
+} OSW_TIME;
+
+typedef struct {
+    uint32_t year;
+    uint32_t month;
+    uint32_t day;
+    uint32_t weekDay;
+    const char* weekDayName;
+} OSW_DATE;
+
 class OswHal {
   public:
     static OswHal* getInstance();
@@ -181,7 +196,7 @@ class OswHal {
     // UTC Time
     void setUTCTime(const time_t& epoch);
     time_t getUTCTime();
-    void getUTCTime(uint32_t* hour, uint32_t* minute, uint32_t* second);
+    void getUTCTime(OSW_TIME* oswTime);
 
     // Offset getters for primary / secondary time (cached!)
     time_t getTimezoneOffsetPrimary();
@@ -189,43 +204,29 @@ class OswHal {
 
     // New time functions with offset
     time_t getTime(time_t& offset);
-    void getTime(time_t& offset, uint32_t* hour, uint32_t* minute, uint32_t* second, bool* afterNoon = nullptr);
-    void getDate(time_t& offset, uint32_t* day, uint32_t* weekDay);
-    void getDate(time_t& offset, uint32_t* day, uint32_t* month, uint32_t* year);
-    const char* getWeekday(time_t& offset, uint32_t* setWDay = nullptr);
+    void getTime(time_t& offset, OSW_TIME* oswTime);
+    void getDate(time_t& offset, OSW_DATE* oswDate, uint32_t* setWDay = nullptr);
 
     // For backward compatibility: Local time functions (= primary timezone)
-    inline void getLocalTime(uint32_t* hour, uint32_t* minute, uint32_t* second, bool* afterNoon = nullptr) {
-        this->getTime(this->timezoneOffsetPrimary, hour, minute, second, afterNoon);
+    inline void getLocalTime(OSW_TIME* oswTime) {
+        this->getTime(this->timezoneOffsetPrimary, oswTime);
     }
     inline uint32_t getLocalTime() {
         return this->getTime(this->timezoneOffsetPrimary);
     }
-    inline void getLocalDate(uint32_t* day, uint32_t* weekDay) {
-        this->getDate(this->timezoneOffsetPrimary, day, weekDay);
-    };
-    inline void getLocalDate(uint32_t* day, uint32_t* month, uint32_t* year) {
-        this->getDate(this->timezoneOffsetPrimary, day, month, year);
-    };
-    inline const char* getLocalWeekday(uint32_t* sWDay = nullptr) {
-        return this->getWeekday(this->timezoneOffsetPrimary, sWDay);
+    inline void getLocalDate(OSW_DATE* oswDate, uint32_t* sWDay = nullptr) {
+        this->getDate(this->timezoneOffsetPrimary, oswDate, sWDay);
     };
 
     // For backward compatibility: Dual time functions (= secondary timezone)
-    inline void getDualTime(uint32_t* hour, uint32_t* minute, uint32_t* second, bool* afterNoon = nullptr) {
-        this->getTime(this->timezoneOffsetSecondary, hour, minute, second, afterNoon);
+    inline void getDualTime(OSW_TIME* oswTime) {
+        this->getTime(this->timezoneOffsetSecondary, oswTime);
     }
     inline uint32_t getDualTime() {
         return this->getTime(this->timezoneOffsetSecondary);
     }
-    inline void getDualDate(uint32_t* day, uint32_t* weekDay) {
-        this->getDate(this->timezoneOffsetSecondary, day, weekDay);
-    };
-    inline void getDualDate(uint32_t* day, uint32_t* month, uint32_t* year) {
-        this->getDate(this->timezoneOffsetSecondary, day, month, year);
-    };
-    inline const char* getDualWeekday(uint32_t* sWDay = nullptr) {
-        return this->getWeekday(this->timezoneOffsetSecondary, sWDay);
+    inline void getDualDate(OSW_DATE* oswDate, uint32_t* sWDay = nullptr) {
+        this->getDate(this->timezoneOffsetPrimary, oswDate, sWDay);
     };
 
     bool _requestDisableBuffer = false;
--- a/src/apps/tools/OswAppDistStats.cpp
+++ b/src/apps/tools/OswAppDistStats.cpp
@@ -15,9 +15,10 @@ void OswAppDistStats::drawChart() {
     uint8_t interval = 20;
     uint16_t goalValue = OswConfigAllKeys::distPerDay.get();
 
-    uint32_t weekDay = 0;
-    uint32_t dayOfMonth = 0;
-    hal->getLocalDate(&dayOfMonth, &weekDay);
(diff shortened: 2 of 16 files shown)


RuffaloLavoisier:
Emulator is working !

simonmicro:
Some thoughts of mine:

* You use `OSW_TIME* oswTime` in the function signatures a lot. I think, as this is mostly the only parameter, it should be required and not optional - so use `OSW_TIME& oswTime` instead.
* Class naming convention. `OSW_TIME` implies a pre-processor `define` or global constant. Maybe use `OswTime` for the `struct`/`class` you are adding.

include/osw_hal.h
```diff
@@ -28,6 +28,21 @@
 #define ERR_SD_MISSING 1
 #define ERR_SD_MOUNT_FAILED 2
 
+typedef struct {
+    uint32_t hour;
+    uint32_t minute;
+    uint32_t second;
+    bool afterNoon;
+} OSW_TIME;
```
simonmicro, line 36: See my comment

include/osw_hal.h
```diff
@@ -28,6 +28,21 @@
...
+
+typedef struct {
+    uint32_t year;
+    uint32_t month;
+    uint32_t day;
+    uint32_t weekDay;
+    const char* weekDayName;
+} OSW_DATE;
```
simonmicro, line 44: See my comment

include/osw_hal.h
```diff
@@ -181,51 +196,37 @@ class OswHal {
     // UTC Time
     void setUTCTime(const time_t& epoch);
     time_t getUTCTime();
-    void getUTCTime(uint32_t* hour, uint32_t* minute, uint32_t* second);
+    void getUTCTime(OSW_TIME* oswTime);
```
simonmicro, line 199: See my comment (also true for all other changed function signatures)

src/apps/tools/OswAppDistStats.cpp
```diff
@@ -15,9 +15,10 @@ void OswAppDistStats::drawChart() {
     uint8_t interval = 20;
     uint16_t goalValue = OswConfigAllKeys::distPerDay.get();
 
-    uint32_t weekDay = 0;
-    uint32_t dayOfMonth = 0;
-    hal->getLocalDate(&dayOfMonth, &weekDay);
+    OSW_DATE oswDate = { 0, };
```
simonmicro, line 18: You are writing C++, so `= {}` is enough.

src/apps/tools/OswAppDistStats.cpp
```diff
@@ -15,9 +15,10 @@ void OswAppDistStats::drawChart() {
...
+    hal->getLocalDate(&oswDate);
+    uint32_t weekDay = oswDate.day;
+    uint32_t dayOfMonth = oswDate.weekDay;
```
simonmicro, line 21: You are making a copy here... Why? Maybe take the reference `&` or just use the value directly.

src/apps/watchfaces/OswAppWatchface.cpp
```diff
@@ -91,16 +94,18 @@ void OswAppWatchface::drawWatch() {
...
     if(OswConfigAllKeys::settingDisplayDualHourTick.get()) {
-        uint32_t dualSecond = 0;
-        uint32_t dualMinute = 0;
-        uint32_t dualHour = 0;
-        hal->getDualTime(&dualHour, &dualMinute, &dualSecond);
+        OSW_TIME oswTime = { 0, };
+        hal->getDualTime(&oswTime);
+        uint32_t dualSecond = oswTime.second;
+        uint32_t dualMinute = oswTime.minute;
+        uint32_t dualHour = oswTime.hour;
```
simonmicro, line 108: see before

src/hal/time.cpp
```diff
@@ -26,37 +26,37 @@ void OswHal::updateTimeProvider() {
...
-void OswHal::getTime(time_t& offset, uint32_t* hour, uint32_t* minute, uint32_t* second, bool* afterNoon) {
+void OswHal::getTime(time_t& offset, OSW_TIME* oswTime) {
     RtcDateTime d = RtcDateTime();
     d.InitWithUnix32Time(this->getTime(offset));
     if (!OswConfigAllKeys::timeFormat.get()) {
         if (d.Hour() > 12) {
-            *hour = d.Hour() - 12;
-            if (afterNoon != nullptr) *afterNoon = true;
+            oswTime->hour = d.Hour() - 12;
+            oswTime->afterNoon = true;
```
simonmicro, line 43: You removed the `if` here - but as the parameter is a pointer, it may be optional... Either change the pointer to force existence (`&`) or condition the assignment with an `if`, please.

src/hal/time.cpp
```diff
@@ -122,27 +122,26 @@ time_t OswHal::getTime(time_t& offset) {
...
+void OswHal::getDate(time_t& offset, OSW_DATE* oswDate, uint32_t* setWDay) {
+    const char* dayMap[7] = {
+        LANG_SUNDAY
+        , LANG_MONDAY
+        , LANG_TUESDAY
+        , LANG_WEDNESDAY
+        , LANG_THURSDAY
+        , LANG_FRIDAY
+        , LANG_SATURDAY
+    };
```
simonmicro, line 134: This is bad. You have an array, allocated on the stack and return a pointer to a value on it. This may be optimized by the compiler, but you are potentially returning a pointer, which becomes invalid with returning from the function / leaving the scope. This issue existed beforehand. Either you make the array `static`, so it will be moved into the `BSS`-section of memory, or you have to find another solution. The current solution is very dangerous...

simonmicro (changes requested):

src/hal/time.cpp
RuffaloLavoisier: Now, it is not pointer and neeed param type not optional.

src/hal/time.cpp
simonmicro: Also, you merge the functions `getDate` and `getWeekday`. But...

* the parameter is still called `setWDay`
* the previous function `getWeekday` allowed you to just get the weekday name OR ask for a specific weekday name - now you always have to provide an `OswTime` (also **nullptr check is missing**) and you can only receive the weekday using this struct

Maybe do not merge these functions?

simonmicro: Also, why did you add the weekday-name to the struct? It is not being used elsewhere (like in `getUTCTime` or `getTime`) and only used here.

simonmicro: The more I think about it... Maybe just remove the whole `getWeekday`-function and potentially replace it with an `const std::array<const char*>`.

include/osw_hal.h
```diff
@@ -181,51 +196,47 @@ class OswHal {
...
-    inline void getDualDate(uint32_t* day, uint32_t* weekDay) {
-        this->getDate(this->timezoneOffsetSecondary, day, weekDay);
+    inline void getDualDate(OswDate& oswDate, uint32_t* sWDay = nullptr) {
+        this->getDate(this->timezoneOffsetPrimary, oswDate, sWDay);
     };
-    inline void getDualDate(uint32_t* day, uint32_t* month, uint32_t* year) {
-        this->getDate(this->timezoneOffsetSecondary, day, month, year);
-    };
-    inline const char* getDualWeekday(uint32_t* sWDay = nullptr) {
-        return this->getWeekday(this->timezoneOffsetSecondary, sWDay);
```
simonmicro, line 228: `For backward compatibility`... I would explicitly not remove these functions for now...

src/apps/tools/OswAppStepStats.cpp
```diff
@@ -43,7 +39,10 @@ void OswAppStepStats::drawInfoPanel(OswUI* ui, uint32_t pos, uint32_t lastWeekDa
     hal->gfx()->setTextCenterAligned();
     hal->gfx()->setTextBottomAligned();
     hal->gfx()->setTextCursor(DISP_W / 2, 170);
-    hal->gfx()->print(hal->getLocalWeekday(&pos));
+    OswDate oswDate = { };
+    hal->getLocalDate(oswDate, &pos);
+    const char* weekday = oswDate.weekDayName;
+    hal->gfx()->print(weekday);
```
simonmicro, line 45: why did you move this down here? I think this code was just a block of drawing instructions before, but now it also contains logic? maybe fence it of into an own `{}` section

src/hal/time.cpp
simonmicro: ...you added the array, but are not using it here?

simonmicro (changes requested):

include/osw_hal.h
RuffaloLavoisier: There is no place to use the old code.

I've already changed the code from the old version to the new one, should I keep it?

simonmicro: Okay... :roll_eyes:

RuffaloLavoisier:
```
[[deprecated("Use OswHal::getDualDate instead.")]]
inline void getDualDate(uint32_t* day, uint32_t* month, uint32_t* year) {
    this->getDate(this->timezoneOffsetSecondary, day, month, year);
};
```
maybe this is style?

simonmicro: Yes - before just removing stuff, first mark it as deprecated. Especially since the compiler will strip unused, deprecated functions anyway...

src/apps/tools/OswAppStepStats.cpp
```diff
@@ -43,7 +39,10 @@ void OswAppStepStats::drawInfoPanel(OswUI* ui, uint32_t pos, uint32_t lastWeekDa
     hal->gfx()->setTextCenterAligned();
     hal->gfx()->setTextBottomAligned();
     hal->gfx()->setTextCursor(DISP_W / 2, 170);
-    hal->gfx()->print(hal->getLocalWeekday(&pos));
+    OswDate oswDate = { };
+    hal->getLocalDate(oswDate);
+    const char* weekday = hal->getWeekDay[oswDate.weekDay];
```
simonmicro, line 44: please use `std::array::at` instead - using the `operator[]` with an invalid index will result in an invalid memory access (as always with C-arrays), but `std::array::at` will throw an exception instead.

simonmicro (approved):

simonmicro:
Hey, the changes to the code itself look fine - but, as expected, LUA is still not compatible with the deprecation attribute. I would recommend you to revert that commit for now and move it into #366, as that PR also aims to fix that.

RuffaloLavoisier:
@simonmicro Ready to merge.

simonmicro (approved):

PR #410, opened 2024-10-09
workflows: fix code style workflows user name

RuffaloLavoisier:
fix code style info -> real uer info


simonmicro:
I'll revert this...

* Because this will assign potentially a lot of additional code changes to only the pipeline triggerer. Example: Multiple deves are working on the same PR. One of them merges it in the end (or better a maintainer of us) and "commits" all the changes.
*This will cause in Git blame to only show him up.
* Also, filtering based on the actions email will not work anymore.
* Furthermore, people signing commits (like me) will cause a lot of unverified commits to pop up.

Issue #414, opened 2024-11-08
Toast message service is not working


RuffaloLavoisier:
# What is not working?

The floating message is not output.

# How to reproduce this

connect BLE . and send message to OSW device.

# What should happen instead?

The floating message must be show on display.

# Environment
- **Model** [e.g. Light Edition] light_edition 3.3
- **OSW-OS** [e.g. Git commit Hash] (830d3054d2076de19aab88a0a9ef2dfa9ebe6c94)
- **Browser** [e.g. Firefox, Chrome] -

# Additional context

simonmicro:
It seems that my machine is working fine, while @RuffaloLavoisier Mac produces a binary with different behavior - even with the same commit hash AND same PlatformIO packages.

Let's see, if the artifacts of the pipeline have the same problem: https://github.com/Open-Smartwatch/open-smartwatch-os/actions/runs/11766491951

simonmicro:
@RuffaloLavoisier Try to clear everything regarding PlatformIO and caches from your system: Remove your `~/.pio` and `~/.platformio` folders, remove the `.pio` folder from the project and perform a clean build. If this still does not help, please upload a firmware, which has this issue here for further analysis.

RuffaloLavoisier:
@simonmicro 

In my side, the same issue happened with macOS workflow..

What do you think is the cause of this issue based on your review? Also, what should the correct values be under normal circumstances?

```
void OswServiceTaskNotifier::loop() {
    const std::lock_guard<std::mutex> lock{mutlimapMutex};
    auto utcTime = std::chrono::system_clock::from_time_t(OswHal::getInstance()->getUTCTime());
    auto currentTime = utcTime + std::chrono::seconds{static_cast<int>(OswHal::getInstance()->getTimezoneOffsetPrimary())};
    if (auto it = scheduler.begin();
            it != scheduler.end() && currentTime >= it->first) { . . . }
}
```

I believe it is a problem if a single piece of code is compiled and runs differently depending on the operating system.

PR #415, opened 2024-11-10
Added notify command to quickly send notifications from console

simonmicro:
...just a quick one to debug #414


simonmicro:
// reverted that from debug, as we already have a `toast` command - lol

PR #420, opened 2024-12-11
workflows: upgrade artifact version to v4

RuffaloLavoisier:
upload, download-artifact plugin v3 will be deprecated, so we upgrade to v4.

### DIFF
--- a/.github/workflows/test-EMULATOR.yml
+++ b/.github/workflows/test-EMULATOR.yml
@@ -56,6 +56,6 @@ jobs:
         if: failure()
         uses: actions/upload-artifact@v4
         with:
-          name: test-results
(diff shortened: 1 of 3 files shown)


RuffaloLavoisier:
See #421

simonmicro (approved):
Btw, what happened to the issue preventing us to use v4 in the first place? The one about artifacts not being available during a run?

RuffaloLavoisier:
@simonmicro 
What exactly was the issue? If I remember correctly, it was an interoperability issue between v3 and v4.
This means the upload and download versions must match exactly.

See https://github.com/Open-Smartwatch/open-smartwatch-os/commit/9fc50051433370103f8fc8e4c2802f40a075194d

PR #423, opened 2025-01-11
A few bugs and minor issues

### LINKED ISSUE #422: Few issues/few enhancement requests
Updating my findings/requests as i go:

# What is not working?

1: whenever watch turns on from sleep to activate screen, it emits a bright white flash before watch face has loaded.

2: Several watch faces that it comes with messes with the font of the app drawer. 

2.1: Watch faces affecting GUI. Large button text until navigating with button, then small overall text:
Analog
Digital
Mix
Dual
Binary
Mono

2.2: Watch faces affecting GUI: 

2.2.1: All text on app drawer is stuck as large text:
Numerals

2.2.2: analog+fit can't be selected. The watch face is frozen, and holding apply button, aka button 2, nothing happens. 

2.2.3: Fitness watch face is too powerful for the ESP32, Holding button 3 to enter app drawer stutters as it increases the bubble size.

3: Wifi connects, but unable to adjust time automatically. Doesn't update.

4: If i set firmware to dd/mm/yyyy as i am used to, watch will not boot up at all after flashing. Just black screen bootloop.

5: If i make adjustment attempts to the code, and re-flash, the welcome screen will not occur.
___
Feature request:

Dedicated setting page to adjust how much force is needed in milligram times sensitivity to activate "wake by wrist movement". As default wakes the watch with the slightest movement.
As well as settings for whether to toggle "wake by wrist movement and wake by tapping"

NEW:
Together with setting page for "wake by lift", add a "automatic calibration submenu so that it can for instance click "calibrate", and it will ask you to flick your wrist 3-5 times to get an average/range of gram force user put into the flick, and save that to the calibration file. And maybe a third page that will output the motion numbers from last calibration, so that users can copy and paste those to a "default user motion data" to a file in the main firmware to auto apply the calibration to future additions if user wants to add/remove apps/functons.

Feature request in firmware to have a toggleable option in firmware.h to make 2 partitions upon flashing. One for the main firmware, second one for battery and accelerometer calibration data.

And make a second toggle for dual partition that enables if "dual partition" option is enabled, which will be a yes/no option to retain current second partition data for battery and accelerometer data. That way you don't need to recalibrate/copy paste new data every time.

Old:

Documentation on default key binds for zooming in/out with buttons, increase/decrease brightness, set watch face as default. As those are missing.
A few links on the readme leads to nowhere/git.io page's 404 error page.

# Environment
- **Model** [e.g. Light Edition]
Light Edition
- **OSW-OS** [e.g. Git commit Hash]
[v0.17.2 /#418
- **Browser** [e.g. Firefox, Chrome]
chrome

# Additional context

simonmicro:
- [x] Fixed select button bug on analog+fitness watchface
- [x] Add note how to use default buttons
- [x] Added "dd/mm/yyyy" as date format
- [x] Toast on default watchface change
- [x] Also reset fonts upon redraw
- [x] Interaction on unlocked console also counts as user interaction
- [x] Reimplemented display+bl init order to prevent flashing

Closes #422 (mostly)

### DIFF
--- a/src/hal/display.cpp
+++ b/src/hal/display.cpp
@@ -64,24 +64,56 @@ bool OswHal::displayBufferEnabled() {
     return this->canvas->hasBuffer();
 }
 
-void OswHal::setupDisplay() {
+void OswHal::setupDisplay(bool fromLightSleep) {
 #ifdef OSW_EMULATOR
     // Always fetch the current instance, just in case the emulator replaced it in the meantime (and as tft is not bound to this objects lifetime)
     tft = fakeDisplayInstance.get();
+#endif
+
+    // Configure backlight pin as early as possible to avoid flickering on startup
+#if OSW_PLATFORM_HARDWARE_DISPLAY_LED != 0
+    ledcAttachPin(OSW_PLATFORM_HARDWARE_DISPLAY_LED, 1);
+    ledcSetup(1, 12000, 8);  // 12 kHz PWM, 8-bit resolution
+    ledcWrite(1, 0); // force off initially
+#else
+#ifndef OSW_EMULATOR // meh, the emulator ignores this for now...
+#warning "Display LED pin unconfigured; can't control backlight brightness"
+#endif
 #endif
 
     // Moved from static allocation to here, as new() operators are limited (size-wise) in that context
     if(!this->canvas)
         this->canvas = new Arduino_Canvas_Graphics2D(DISP_W, DISP_H, tft);
 
-    this->canvas->begin();  // use default speed and default SPI mode
+    if(!fromLightSleep) {
+        this->canvas->begin();  // will not deconfigured upon light sleep, use default speed and default SPI mode
 
-    // Another nasty hack to avoid display flicker when turning on
-    // Let the display settle a bit befor sending commands to it.
-    // This avoids a partially white screen on turning on the device.
-    // 45 ms is tested on an V3.3 device.
-    delay(45);
+        // Clean the buffer, if it is enabled (otherwise it would take too long during system startup)
+        if(this->displayBufferEnabled()) {
+            this->gfx()->fillBuffer(rgb565(0, 0, 0));
+            this->flushCanvas();
+        }
+    }
+
+    // Finally, enable the backlight and show the canvas
     this->displayOn();
+}
+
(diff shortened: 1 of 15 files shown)


src/hal/display.cpp
```diff
@@ -64,26 +64,58 @@ bool OswHal::displayBufferEnabled() {
...
+        this->canvas->begin();  // will not deconfigured upon light sleep, use default speed and default SPI mode
 
-    // Another nasty hack to avoid display flicker when turning on
-    // Let the display settle a bit befor sending commands to it.
-    // This avoids a partially white screen on turning on the device.
-    // 45 ms is tested on an V3.3 device.
-    delay(45);
+        // Clean the buffer, if it is enabled (otherwise it would take too long during system startup)
+        if(this->displayBufferEnabled()) {
+            this->gfx()->fillBuffer(rgb565(0, 0, 0));
```
RuffaloLavoisier, line 93: How about `ui->getBackgroundColor()`?

src/hal/display.cpp
```diff
@@ -64,26 +64,58 @@ bool OswHal::displayBufferEnabled() {
...
+    // Finally, enable the backlight and show the canvas
     this->displayOn();
 }
 
+void OswHal::stopDisplay(bool toLightSleep) {
+    this->displayOff();
+
+    // If we have a fast buffer, also clean it, to prevent flashing upon returning from light sleep (deep sleep will loose the buffer anyway, but as this is fast...)
+    if(!toLightSleep and this->displayBufferEnabled()) {
+        this->gfx()->fillBuffer(rgb565(0, 0, 0));
```
RuffaloLavoisier, line 107: How about `ui->getBackgroundColor()`?

RuffaloLavoisier (changes requested):

src/hal/display.cpp
simonmicro: Background color maybe white and this will once again lead to flashing upon start. Using black, a user may not even notice the backlight flickering on.

src/hal/display.cpp
simonmicro: Background color maybe white and this will once again lead to flashing upon start. Using black, a user may not even notice the backlight flickering on.

RuffaloLavoisier (approved):
LGTM

PR #409, opened 2024-10-08
Create StepCount Service Server

RuffaloLavoisier:
Manage From StepCount API.

Issue #141, opened 2021-08-04
Open-smartwatch application for Smartphone


RuffaloLavoisier:
Hi, I got a lot of inspiration from my watch.  You can do more various things by linking your wristwatch and smartphone with ESP32. I have a similar function.

![oledwatch](https://user-images.githubusercontent.com/66334969/128190448-b6c96259-5bf2-4780-b381-a8bc912888f1.jpg)

# 1. Notification
It informs the alarm that comes to the smartphone with a clock.

# 2. Device Finder
Default : If they are out of a certain distance from each other, the wristwatch gives an alarm to inform the user of the risk of loss. (This is usually the case when you are looking for a smartphone. )

## from phone
When a finder function is activated on a smartphone, the location of the clock can be checked through an alarm built into the clock. 

## from watch
Select the mode from the clock and make the smartphone ring to locate it. 

Is there anyone who wants to contribute to this?

kekko7072:
Hey I am a mobile app developer with flutter and I work full stack (server + front end), but I am not very skilled at electronics but I want to improve. I had your same idea, and it is exiting. I've abandoned the idea because my skill using and editing the firmware of the watch were very poor, but if you help me with the hardware/software of the watch I will be glad to develop the app. Last year I bought a watch light edition but I was not able to edit the firmware and run bluetooth.

RuffaloLavoisier:
@kekko7072 
Hi, there
Is it possible to build OSW firmware now? Although there is not enough memory to use Bluetooth, it has been confirmed that it can be implemented. Personally, if you have a prototype, the pull request is always open.Please feel free to open a discussion

kekko7072:
Hey @RuffaloLavoisier what do you mean for OSW?

RuffaloLavoisier:
Sorry, `Open-smartwatch` build !

simonmicro:
-> https://github.com/Open-Smartwatch/open-smartwatch-os/blob/bc08af49ac33bde2c81f463ee0a1be0812b8d388/src/services/OswServiceTaskWebserver.cpp#L291

PierreBresson:
Hi,

[...]

PR #424, opened 2025-01-11
Implemented basic BLE server

simonmicro:
After #409, part of #141 

this is JUST the server, so a device can connect to an OSW and perform a few basic operations

the client will be implemented whenever we have a spec/server for it

checklist:
- [x] config option to start server on boot
- [x] document new flags in docs
- [x] basic ui for
  - [x] binding request
  - [x] connect/disconnect
  - [x] toasts
- [x] services
  - [x] firmware version
  - [x] battery info
  - [x] custom OSW service

### DIFF
--- a/include/services/OswServiceManager.h
+++ b/include/services/OswServiceManager.h
@@ -25,7 +25,12 @@ class OswServiceManager {
 #else
 #define _SERVICE_WIFI 0
 #endif
-    const unsigned workerStackSize = 2048 + (8192 * _SERVICE_WIFI); // If wifi is active, set to same size as core 0
+#ifdef OSW_FEATURE_BLE_SERVER
+#define _SERVICE_BLE 1
+#else
+#define _SERVICE_BLE 0
+#endif
+    const unsigned workerStackSize = 2048 + (8192 * _SERVICE_WIFI) + (2048 * _SERVICE_BLE); // If wifi is active, set to same size as core 0
     const unsigned workerStartupDelay = 2000;
     const unsigned workerLoopDelay = 10;
 
--- a/include/services/OswServiceManager.h
+++ b/include/services/OswServiceManager.h
@@ -19,18 +19,18 @@ class OswServiceManager {
         instance.reset();
     }
 
-    //Temp workaround until #137 is done
+    // Temp workaround until #137 is done
 #ifdef OSW_FEATURE_WIFI
-#define _SERVICE_WIFI 1
+#define _ADDITIONAL_STACK_SIZE_FOR_WIFI 8192
 #else
-#define _SERVICE_WIFI 0
+#define _ADDITIONAL_STACK_SIZE_FOR_WIFI 0
 #endif
 #ifdef OSW_FEATURE_BLE_SERVER
-#define _SERVICE_BLE 1
+#define _ADDITIONAL_STACK_SIZE_FOR_BLE 2048
 #else
-#define _SERVICE_BLE 0
+#define _ADDITIONAL_STACK_SIZE_FOR_BLE 0
 #endif
-    const unsigned workerStackSize = 2048 + (8192 * _SERVICE_WIFI) + (2048 * _SERVICE_BLE); // If wifi is active, set to same size as core 0
+    const unsigned workerStackSize = 2048 + _ADDITIONAL_STACK_SIZE_FOR_WIFI + _ADDITIONAL_STACK_SIZE_FOR_BLE; // base stack size is 2k
     const unsigned workerStartupDelay = 2000;
     const unsigned workerLoopDelay = 10;
 
--- a/src/osw_ui.cpp
+++ b/src/osw_ui.cpp
@@ -168,10 +168,10 @@ void OswUI::loop() {
         {
             std::lock_guard<std::mutex> notifyGuard(this->mNotificationsLock);
             // Draw all notifications
-            auto y = DISP_H - OswUINotification::sDrawHeight;
+            auto y = DISP_H;
             for (const auto& notification : this->mNotifications) {
+                y -= notification.getDrawHeight();
                 notification.draw(y);
-                y -= OswUINotification::sDrawHeight;
             }
         }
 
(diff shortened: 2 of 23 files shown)


src/services/OswServiceTaskBLEServer.cpp
```diff
@@ -301,25 +301,19 @@ void OswServiceTaskBLEServer::BatteryLevelStatusCharacteristicCallbacks::onRead(
...
-    this->bytesCurrentTime[2] = (uint8_t) month; // Exact Time 256 -> Day Date Time -> Date Time -> Month
-    this->bytesCurrentTime[3] = (uint8_t) day; // Exact Time 256 -> Day Date Time -> Date Time -> Day
-    this->bytesCurrentTime[4] = (uint8_t) hour; // Exact Time 256 -> Day Date Time -> Date Time -> Hours
-    this->bytesCurrentTime[5] = (uint8_t) minute; // Exact Time 256 -> Day Date Time -> Date Time -> Minutes
-    this->bytesCurrentTime[6] = (uint8_t) second; // Exact Time 256 -> Day Date Time -> Date Time -> Seconds
-    this->bytesCurrentTime[7] = (uint8_t) (dow == 0 ? 7 : dow); // Exact Time 256 -> Day Date Time -> Day of Week
+    OswDate oswDate;
+    OswHal::getInstance()->getDate(offset, oswDate);
+    OswTime oswTime;
+    OswHal::getInstance()->getUTCTime(oswTime);
```
RuffaloLavoisier, line 307: why not use `getLocalDate/getLocalTime`?

src/services/OswServiceTaskBLEServer.cpp
```diff
@@ -301,25 +301,19 @@ void OswServiceTaskBLEServer::BatteryLevelStatusCharacteristicCallbacks::onRead(
 
 void OswServiceTaskBLEServer::CurrentTimeCharacteristicCallbacks::onRead(NimBLECharacteristic* pCharacteristic) {
     time_t offset = 0;
```
RuffaloLavoisier, line 303: ?

RuffaloLavoisier: why are you use offset and init to `0`.

RuffaloLavoisier (changes requested):

src/services/OswServiceTaskBLEServer.cpp
simonmicro: If you have such questions, please take a look into the specification first: https://www.bluetooth.com/specifications/specs/current-time-service-1-1/

-> We implement the current time service with the current time field, NOT local time field.

src/services/OswServiceTaskBLEServer.cpp
simonmicro: Because we have no function called `getUTCDate` - we have only functions for local dates, or the UTC time. Hence, I needed to provide the offset for this. But you are right, I've changed the time getter to match the pattern too.

include/services/OswServiceManager.h
```diff
@@ -25,7 +25,12 @@ class OswServiceManager {
 #else
 #define _SERVICE_WIFI 0
 #endif
-    const unsigned workerStackSize = 2048 + (8192 * _SERVICE_WIFI); // If wifi is active, set to same size as core 0
+#ifdef OSW_FEATURE_BLE_SERVER
+#define _SERVICE_BLE 1
+#else
+#define _SERVICE_BLE 0
+#endif
+    const unsigned workerStackSize = 2048 + (8192 * _SERVICE_WIFI) + (2048 * _SERVICE_BLE); // If wifi is active, set to same size as core 0
```
RuffaloLavoisier, line 33: can we use macro? I want use value from macro.

like

```
#define WHAT_IS_MEAN 2048 // option: comment
```

src/osw_ui.cpp
```diff
@@ -160,10 +163,10 @@ void OswUI::loop() {
         {
             std::lock_guard<std::mutex> notifyGuard(this->mNotificationsLock);
             // Draw all notifications
-            auto y = DISP_H - OswUINotification::sDrawHeight;
+            auto y = DISP_H;
```
RuffaloLavoisier, line 166: please rename `y`.

src/osw_ui.cpp
```diff
@@ -298,20 +304,44 @@ void OswUI::OswUIProgress::draw() {
...
+
 void OswUI::OswUINotification::draw(unsigned y) const {
-    // TODO
-    // * handle too long texts by adding a scroll animation?
-    // * newlines may change a notifications height, so the y position should be calculated based on the current height of the notification -> return that instead of void
+    // TODO handle too long texts by adding a scroll animation?
     Graphics2DPrint* gfx = OswHal::getInstance()->gfx();
-    gfx->fillFrame(0, y, DISP_W, this->sDrawHeight, OswUI::getInstance()->getBackgroundColor());
+    auto height = this->getDrawHeight();
+    gfx->fillFrame(0, y, DISP_W, height, OswUI::getInstance()->getBackgroundColor());
```
RuffaloLavoisier, line 340: same, rename `y`.

src/services/OswServiceTaskConsole.cpp
```diff
@@ -78,11 +91,57 @@ void OswServiceTaskConsole::runPrompt() {
...
+#ifndef OSW_EMULATOR
+        } else if (this->m_inputBuffer == "reboot") {
+            // this does not work in the emulator as it is running under an own thread, of which the shutdown-exception is not captured - populating here and crashing
+            ESP.restart();
+#endif
+#if OSW_SERVICE_NOTIFIER == 1
+        } else if (this->m_inputBuffer.find("toast ") == 0 and this->m_inputBuffer.length() > 6) {
+            auto arg = this->m_inputBuffer.substr(6);
+            std::string toast;
+            for (auto i = 0; i < arg.length(); i++) {
```
RuffaloLavoisier, line 121: rename `i`,

src/osw_ui.cpp
```diff
@@ -298,20 +304,44 @@ void OswUI::OswUIProgress::draw() {
 size_t OswUI::OswUINotification::count{};
 
 OswUI::OswUINotification::OswUINotification(std::string message, bool isPersistent)
-    : message{message.c_str()}, endTime{isPersistent ? millis() + 300'000 : millis() + 5'000} {
-    id = count;
+    : id(count), message{message.c_str()}, lines(countLines(message)), endTime{millis() + (isPersistent ? 300'000 : 5'000) * lines} {
```
RuffaloLavoisier, line 307: change to macro.

src/osw_ui.cpp
```diff
@@ -298,20 +304,44 @@ void OswUI::OswUIProgress::draw() {
...
+    return lines;
+}
+
+/**
+ * @brief Get the draw height based on line count with additional padding
+ *
+ * @return unsigned char
+ */
+unsigned char OswUI::OswUINotification::getDrawHeight() const {
+    return 4 + 8 * lines + 4; // padding, line height, padding
```
RuffaloLavoisier, line 333: Also....maybe?

RuffaloLavoisier (changes requested):

src/osw_ui.cpp
simonmicro: What name would you propose? It _is_ literally the y coordinate being used here. Additionally, as this scope is only 5 lines long... What would be a better name?

src/osw_ui.cpp
simonmicro: Same issue here... `yStart`? `posY`? Idk, take a look into the GFX2D library, they also use one-letter names for coordinate values there.

src/osw_ui.cpp
RuffaloLavoisier: `y` It's simple, but I think it's best if it's a clear and instantly identifiable name that can be read without thinking in the distant future or when others see it.

src/services/OswServiceTaskConsole.cpp
simonmicro: It is the index... What do you propose? I think using something like `argIdx` (this even sounds misleading), `argumentCharacterIndex` or `index` does not yield additional value, especially since this `for`-loop is so short...

RuffaloLavoisier: IMHO - Sometimes we write simple and similar names for simple variables within a short range, but long codes seem to confuse.

src/osw_ui.cpp
RuffaloLavoisier: Using variable names like x and y is simple and intuitive, especially when they represent coordinates such as the x-axis and y-axis. However, if they appear in many places within the code, it can become harder to understand the overall context. 

Context Dependence: The meaning of x and y may only be clear from the surrounding context, making the code harder to read.
Potential Confusion: In more complex codebases, x and y might not always clearly convey their purpose.
Lack of Scalability: If additional coordinates or variables are introduced, temporary names like x1 or y1 may appear, reducing clarity.

But this case, It doesn't seem to have much effect on the current code, so looks good to me.

RuffaloLavoisier (approved):

PR #428, opened 2025-01-12
Fix compile minor waring

RuffaloLavoisier:
Move variables that are compiled but may not be used due to macros into the scope of the macros.


simonmicro (approved):
at some point we should find a better solution to this

PR #441, opened 2026-05-22
 Fix OswUI::stopProgress() never freeing the progress bar

RuffaloLavoisier:
Invert condition in stopProgress to check inactive state


simonmicro (approved):
ooops

RuffaloLavoisier:
Merged, Thanks !

PR #443, opened 2026-06-13
feat: support building the emulator on macOS (incl. Apple Silicon)

RuffaloLavoisier:
# BIG NEWS - WE CAN SUPPORT macOS BUILD !

The emulator was only validated on Linux (GCC/libstdc++). This makes it build and run on macOS (Apple clang/libc++) while keeping Linux unchanged: all OS-specific bits are gated behind the existing `Darwin` branch, and the source-level changes are portable standard C++ that behave identically on both toolchains.

Changes:
- Rename the Arduino String shim `emulator/include/String.h` to `WString.h` and delete the old file. On macOS's case-insensitive filesystem `String.h` shadowed the C standard `<string.h>`, breaking libc++ internals (std::memcmp / __murmur2_or_cityhash). Repoint all includers (FakeMe.h, Preferences.h, String.cpp, unit tests) to WString.h.
- CMake (Darwin branch only): enable `_LIBCPP_ENABLE_EXPERIMENTAL` so std::jthread is available (Apple libc++ keeps it experimental; libstdc++ ships it by default).
- CMake (Darwin branch only): replace the dead, Intel-only `link_directories(/usr/local/lib)` with `target_link_directories(... ${SDL2_LIBRARY_DIRS} ${SDL2IMAGE_LIBRARY_DIRS})` so Homebrew's SDL2 is found on Apple Silicon (/opt/homebrew/lib).
- Unit tests: replace the GCC-only `42.042d` literal with standard `42.042`, and disambiguate `char + String` via an explicit cast to std::string. Both compile and behave the same under GCC.

Note: the 3 `emulator.run_headless*` unit tests fail on macOS because the test fixture runs the emulator on a background thread, while macOS Cocoa requires SDL event pumping on the main thread (NSException). This is a test-harness limitation only — running the emulator normally (main thread) boots the full OS cleanly. 35/39 tests pass, 1 skipped by design.

image


simonmicro:
So... This seems to be a subset of the already existing https://github.com/Open-Smartwatch/open-smartwatch-os/pull/436 . I really discourage this practice and will close this one as such.

I'll take a look at the existing PR instead.

